PR #16, opened 2019-12-27
support sorting output by primary key

tangenta:
Support sorting output by primary key.
Use [DATA-DOG/go-sqlmock](https://github.com/DATA-DOG/go-sqlmock) to mock `database/sql` related types.

### DIFF
--- a/cmd/dumpling/main.go
+++ b/cmd/dumpling/main.go
@@ -57,11 +57,13 @@ func init() {
 	flag.Uint64Var(&fileSize, "F", export.UnspecifiedSize, "The approximate size of output file")
 	flag.Uint64Var(&fileSize, "filesize", export.UnspecifiedSize, "The approximate size of output file")
 
-	flag.StringVar(&outputDir, "output", defaultOutputDir(), "Output directory")
-	flag.StringVar(&outputDir, "o", defaultOutputDir(), "Output directory")
+	flag.StringVar(&outputDir, "output", defaultOutputDir, "Output directory")
+	flag.StringVar(&outputDir, "o", defaultOutputDir, "Output directory")
 }
 
-func defaultOutputDir() string {
+var defaultOutputDir = timestampDirName()
+
+func timestampDirName() string {
 	return fmt.Sprintf("./export-%s", time.Now().Format(time.RFC3339))
 }
 
--- a/cmd/dumpling/main.go
+++ b/cmd/dumpling/main.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	_ "net/http/pprof"
 	"os"
+	"time"
 
 	"github.com/pingcap/dumpling/v4/cli"
 	"github.com/pingcap/dumpling/v4/export"
@@ -56,8 +57,12 @@ func init() {
 	flag.Uint64Var(&fileSize, "F", export.UnspecifiedSize, "The approximate size of output file")
 	flag.Uint64Var(&fileSize, "filesize", export.UnspecifiedSize, "The approximate size of output file")
 
-	flag.StringVar(&outputDir, "output", ".", "Output directory")
-	flag.StringVar(&outputDir, "o", ".", "Output directory")
+	flag.StringVar(&outputDir, "output", defaultOutputDir(), "Output directory")
+	flag.StringVar(&outputDir, "o", defaultOutputDir(), "Output directory")
+}
+
+func defaultOutputDir() string {
+	return fmt.Sprintf("./export-%s", time.Now().Format(time.RFC3339))
 }
 
 func main() {
--- a/v4/export/config.go
+++ b/v4/export/config.go
@@ -3,8 +3,9 @@ package export
 import (
 	"fmt"
 	"regexp"
-	"strconv"
 	"strings"
+
+	"github.com/coreos/go-semver/semver"
 )
 
 type Config struct {
@@ -33,7 +34,7 @@ func DefaultConfig() *Config {
 		Logger:        &DummyLogger{},
 		FileSize:      UnspecifiedSize,
 		OutputDirPath: ".",
-		ServerInfo:    UnknownServerInfo,
+		ServerInfo:    ServerInfoUnknown,
 		SortByPk:      false,
 	}
 }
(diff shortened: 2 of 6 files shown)


cmd/dumpling/main.go
```diff
@@ -56,8 +57,12 @@ func init() {
 	flag.Uint64Var(&fileSize, "F", export.UnspecifiedSize, "The approximate size of output file")
 	flag.Uint64Var(&fileSize, "filesize", export.UnspecifiedSize, "The approximate size of output file")
 
-	flag.StringVar(&outputDir, "output", ".", "Output directory")
-	flag.StringVar(&outputDir, "o", ".", "Output directory")
+	flag.StringVar(&outputDir, "output", defaultOutputDir(), "Output directory")
+	flag.StringVar(&outputDir, "o", defaultOutputDir(), "Output directory")
```
kennytm: Save the `defaultOutputDir()` into a variable so that both `-o` and `--output` are using the same default value.

v4/export/config.go
```diff
@@ -28,28 +33,85 @@ func DefaultConfig() *Config {
...
+	ServerType    ServerType
+	ServerVersion *ServerVersion
+}
+
+var UnknownServerInfo = ServerInfo{
+	ServerType:    UnknownServerType,
+	ServerVersion: nil,
+}
+
+var versionRegex = regexp.MustCompile("^(\\d+\\.){2}\\d+")
```
kennytm:
```suggestion
var versionRegex = regexp.MustCompile(`^\d+\.\d+\.\d+`)
```

v4/export/config.go
```diff
@@ -28,28 +33,85 @@ func DefaultConfig() *Config {
...
+var tidbVersionRegex = regexp.MustCompile("v(\\d+\\.){2}\\d+")
```
kennytm: For TiDB, if the server reports `v3.1.0-beta.1`, that `-beta.1` part is also important I think.

v4/export/config.go
```diff
@@ -28,28 +33,85 @@ func DefaultConfig() *Config {
...
-func extractOutputConfig(conf *Config) *Config {
-	return &Config{
-		Logger:        conf.Logger,
-		FileSize:      conf.FileSize,
-		OutputDirPath: conf.OutputDirPath,
+const UnspecifiedSize = 0
+
+type ServerInfo struct {
+	ServerType    ServerType
+	ServerVersion *ServerVersion
```
kennytm: Let's use a semver library to store the server version (TiDB uses `github.com/coreos/go-semver`). So we don't need to repeat that "parse `1.2.3` into `[1, 2, 3]`" code.

v4/export/config.go
```diff
@@ -28,28 +33,85 @@ func DefaultConfig() *Config {
...
-	// Logger is used to log the export routine.
-	Logger Logger
-	// Output size limit in bytes.
-	OutputSize int
-	// OutputDirPath is the directory to output.
-	OutputDirPath string
+type ServerType int8
+
+const (
+	UnknownServerType = iota
```
kennytm:
```suggestion
	UnknownServerType ServerType = iota
```

BTW better name the constants `ServerTypeUnknown` etc for better auto-completion behavior.

v4/export/dump.go
```diff
@@ -206,3 +237,59 @@ func dumpTable(ctx context.Context, db *sql.DB, database, table string) (TableDa
...
+		return fmt.Sprintf("ORDER BY %s", pkName), nil
+	}
+	switch conf.ServerInfo.ServerType {
+	case TiDBServerType:
+		return "ORDER BY _tidb_rowid", nil
+	default:
+		// ignore when there is no primary key
+		return "", nil
+	}
+}
```
kennytm: In TiDB, please always `ORDER BY _tidb_rowid` unless PKIsHandle. The default behavior without any ORDER BY clause should be sufficient. 

The purpose of "order by primary key" is to ensure the KV pairs generated are ordered, so overlapping data range will happen less. The ORDER BY clause shouldn't slow down the dumping speed in normal cases.

v4/export/dump.go
```diff
@@ -206,3 +237,59 @@ func dumpTable(ctx context.Context, db *sql.DB, database, table string) (TableDa
...
+
+const PrimaryKeyQuery = `SELECT COLUMN_NAME
+	FROM INFORMATION_SCHEMA.COLUMNS
+	WHERE TABLE_SCHEMA = '%s'
+	AND TABLE_NAME = '%s'
+	AND COLUMN_KEY = 'PRI';`
```
kennytm: Please use prepared statement for this, instead of `Sprintf`'ing the names. The table names can contain a `'`.

```sql
SELECT column_name FROM information_schema.columns
WHERE table_schema = ? AND table_name = ? AND column_key = 'PRI';
```

v4/export/dump_test.go
```diff
@@ -0,0 +1,111 @@
...
+		rowsMaker, result := n[3].(func() *sqlmock.Rows), n[5].(string)
+
+		mockConf.SortByPk = orderByPk
+		if mockConf.SortByPk {
+			mock.ExpectQuery(getPkQuery).WillReturnRows(rowsMaker())
+		}
+		if isTiDBType {
+			mockConf.ServerInfo.ServerType = TiDBServerType
+		} else {
+			mockConf.ServerInfo.ServerType = randDBType([]ServerType{UnknownServerType, MySQLServerType, MariaDBServerType})
```
kennytm: Why not just test all 4 server types?  Tests should best be comprehensive and deterministic. There's no reason using `rand` here.

v4/export/dump.go
tangenta: Is there a convenient way to know whether a table is PKIsHandle?

As far as I know, "a table contains the primary key" cannot infer "PKIsHandle", because if the config option `alter-primary-key` is enabled, all new-created tables are PKIsNotHandle.
See: https://github.com/pingcap/tidb/pull/13229

Can we simply try to `select _tidb_rowid from t` to see if it returns an error?

kennytm: @tangenta 

> Can we simply try to `select _tidb_rowid from t` to see if it returns an error?

Yes, that's how mydumper checks if `_tidb_rowid` exists.

Or, if we trust TiDB to be always ordered by PK, just don't include the ORDER BY 🙃 (the checklist is most relevant when we dump single table in parallel).

v4/export/dump.go
```diff
@@ -206,3 +234,82 @@ func dumpTable(ctx context.Context, db *sql.DB, database, table string) (TableDa
...
+		}
+	}
+	return query.String(), nil
+}
+
+const errBadFieldCode = 1054
+
+func buildOrderByClause(conf *Config, db *sql.DB, database, table string) (string, error) {
+	if conf.ServerInfo.ServerType == ServerTypeTiDB {
+		tiDBRowIDQuery := fmt.Sprintf("SELECT _tidb_rowid from %s.%s", database, table)
```
kennytm:
```suggestion
		tiDBRowIDQuery := fmt.Sprintf("SELECT _tidb_rowid from %s.%s LIMIT 0", database, table)
```

PR #19, opened 2019-12-29
support dumping binary data in hexadecimal representation

tangenta:


v4/export/ir.go
```diff
@@ -19,6 +21,15 @@ type SQLRowIter interface {
...
 }
 
+type RowReceiverStringer interface {
+	RowReceiver
+	Stringer
+}
+
+type Stringer interface {
+	ToString() string
+}
```
kennytm: why not reuse `fmt.Stringer`?

v4/export/test_util.go
```diff
@@ -71,65 +74,67 @@ func newMockMetaIR(targetName string, meta string, specialComments []string) Met
...
-		idx:  0,
-		data: m.data,
+func newMockTableIR(databaseName, tableName string, data [][]interface{}, specialComments, colTypes []string) TableDataIR {
+	values := make([][]driver.Value, len(data))
+	for i := range values {
+		values[i] = make([]driver.Value, len(data[i]))
+		for j := range data[i] {
+			values[i][j] = data[i][j]
+		}
+	}
```
kennytm: could `values := data` work? 🤔

v4/export/writer_util.go
```diff
@@ -108,37 +86,127 @@ func wrapStringWith(str string, wrapper string) string {
...
+	"DATE":      SQLTypeStringMaker,
+	"TIME":      SQLTypeStringMaker,
+	"YEAR":      SQLTypeStringMaker,
+
+	"BINARY":    SQLTypeBytesMaker,
+	"VARBINARY": SQLTypeBytesMaker,
+	"BLOB":      SQLTypeBytesMaker,
+	"BIT":       SQLTypeBytesMaker,
+
+	"INT":     SQLTypeIntegerMaker,
```
kennytm: Can this handle MEDIUMBLOB / BIGINT / etc?

v4/export/writer_util.go
```diff
@@ -108,37 +86,127 @@ func wrapStringWith(str string, wrapper string) string {
...
+}
+func (s *SQLTypeString) ReportSize() uint64 {
+	if s.Valid {
+		return uint64(len(s.String))
+	}
+	return uint64(len("NULL"))
+}
+func (s *SQLTypeString) ToString() string {
+	if s.Valid {
+		return fmt.Sprintf(`'%s'`, s.String)
```
kennytm: What if the string contains a `'` (e.g. `'this shouldn''t fail'`)

v4/export/writer_util.go
```diff
@@ -108,37 +86,127 @@ func wrapStringWith(str string, wrapper string) string {
...
+	"TIME":      SQLTypeStringMaker,
+	"YEAR":      SQLTypeStringMaker,
+
+	"BINARY":    SQLTypeBytesMaker,
+	"VARBINARY": SQLTypeBytesMaker,
+	"BLOB":      SQLTypeBytesMaker,
+	"BIT":       SQLTypeBytesMaker,
+
+	"INT":     SQLTypeIntegerMaker,
+	"FLOAT":   SQLTypeIntegerMaker,
```
kennytm: FLOAT is not an "integer" 🙃

v4/export/ir.go
tangenta: There is a naming collide between `fmt.Stringer`'s `String()` and `sql.NullString`'s `String` field.

v4/export/test_util.go
tangenta:
```
cannot use values (type [][]interface {}) as type [][]driver.Value in field value
```

v4/export/writer_util.go
tangenta: Sure, we can add fill this map gradually.

v4/export/test_util.go
kennytm: Could `data` be `[][]driver.Value` instead of `[][]interface{}`?

v4/export/writer_util.go
```diff
@@ -191,12 +191,16 @@ func (s *SQLTypeString) ReportSize() uint64 {
...
 	if s.Valid {
-		return fmt.Sprintf(`'%s'`, s.String)
+		return fmt.Sprintf(`'%s'`, escapeSingleQuote(s.String))
 	} else {
 		return "NULL"
 	}
 }
 
+func escapeSingleQuote(src string) string {
+	return strings.ReplaceAll(src, "'", "\\'")
```
kennytm: The SQL-compliant escaping is `''` not `\'`. 

And if we need to support `\'` we gotta escape `\` as `\\` too.

v4/export/writer_util.go
kennytm: Ok.

PR #20, opened 2019-12-30
add support for consistency control

tangenta:
Changes:
- refactor `Dump()`: we first list all the tables, then start dumping.
- add tests for `dumpTable()`.
- add three kinds of consistency control: 
  - lock all tables
  - flush tables with read lock
  - snapshot
- add some SQL convenient functions.

### DIFF
--- a/v4/export/dump.go
+++ b/v4/export/dump.go
@@ -3,7 +3,7 @@ package export
 import (
 	"context"
 	"database/sql"
-	"sync"
+	"golang.org/x/sync/errgroup"
 
 	_ "github.com/go-sql-driver/mysql"
 )
@@ -30,9 +30,8 @@ func Dump(conf *Config) (err error) {
 		return err
 	}
 
-	conCtrl := ConsistencyController{}
-	err = conCtrl.Setup(conf, ConsistencyNone)
-	if err != nil {
+	var conCtrl ConsistencyController = &ConsistencyNone{}
+	if err = conCtrl.Setup(); err != nil {
 		return err
 	}
 
@@ -40,10 +39,10 @@ func Dump(conf *Config) (err error) {
 	if err != nil {
 		return err
 	}
-
 	if err = dumpDatabases(context.Background(), conf, pool, fsWriter); err != nil {
 		return err
 	}
+
 	return conCtrl.TearDown()
 }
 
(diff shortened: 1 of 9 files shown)


v4/export/dump.go
```diff
@@ -3,112 +3,100 @@ package export
...
+			}(i, table, &wg, res)
+		}
+		wg.Wait()
+		for _, err := range res {
 			if err != nil {
-				res[ith] = err
-				return
+				return err
 			}
+		}
```
kennytm: Could we use [errgroup](https://godoc.org/golang.org/x/sync/errgroup) for this?

BTW this will dump all tables even if one of them failed (late exit), while `errgroup` will stop as soon as one of the table failed (early exit). Not sure if which is the desired behavior.

v4/export/prepare.go
```diff
@@ -0,0 +1,160 @@
...
+	case ConsistencyFlushTableWithReadLock:
+		return c.setupFlushTableWithReadLock()
+	case ConsistencyLockDumpingTables:
+		return c.setupTableLock()
+	case ConsistencySnapshot:
+		return c.setupSnapshot()
+	default:
+		panic("unsupported consistency strategy")
+	}
+	return nil
```
kennytm:
```suggestion
```

This `return nil` should never be reachable.

v4/export/prepare.go
```diff
@@ -0,0 +1,160 @@
...
+	case ConsistencyFlushTableWithReadLock:
+		return c.tearDownFlushTableWithReadLock()
+	case ConsistencyLockDumpingTables:
+		return c.tearDownTableLock()
+	case ConsistencySnapshot:
+		return c.tearDownSnapshot()
+	default:
+		panic("unsupported consistency strategy")
+	}
+	return nil
```
kennytm:
```suggestion
```

v4/export/prepare.go
```diff
@@ -0,0 +1,160 @@
...
+		tables, err := ShowTables(db)
+		if err != nil {
+			return nil, err
+		}
+		dbTables[dbName] = tables
+	}
+	return dbTables, nil
+}
+
+type Consistency int8
```
kennytm: IMO `Consistency` looks more like an `interface` than an enum (in particular you're calling it a "strategy") 🤷.

```go
type Consistency interface {
    setupConsistency(*ConsistencyController) error
    tearDownConsistency(*ConsistencyController) error
}
```

v4/export/sql.go
```diff
@@ -172,6 +172,47 @@ func GetPrimaryKeyName(db *sql.DB, database, table string) (string, error) {
...
+	}
+	err := simpleQuery(db, "SHOW MASTER STATUS", handleOneRow)
+	if err != nil {
+		return nil, err
+	}
+	return oneRow, nil
+}
+
+func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
+	_, err := db.Exec(fmt.Sprintf("SET SESSION tidb_snapshot = '%s'", snapshot))
```
kennytm:
```suggestion
	_, err := db.Exec("SET SESSION tidb_snapshot = ?", snapshot)
```

PR #21, opened 2020-01-06
add integration test framework and naughty strings test case

tangenta:
Add:
- MySQL server in travis CI
- an integration test framework
- an integration  test case for naughty strings


kennytm:
Could we write the integration tests in a scripting language like Bash or Python (see BR and Lightning)? It's easier to contribute than Go. 

I'd expect the test to be simply running

```sh
set -e

mysql < input.sql
bin/dumpling "...flags..." -o output.sql
diff input.sql output.sql
```

PR #23, opened 2020-01-08
log: replace the old logger with `zap`

tangenta:


v4/log/log.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"log"
+
+	"go.uber.org/zap"
+)
+
+var globalLogger *zap.Logger
+
+func init() {
+	logger, err := zap.NewDevelopment(zap.AddCaller())
```
kennytm: should we wrap `pingcap/log` for consistency with other projects?

PR #37, opened 2020-03-04
tests: some improvements to the integration tests

kennytm:
1. Place the password into a config file when executing `mysql`. This avoid the constant "Using a password on the command line interface can be insecure" warning (see https://stackoverflow.com/questions/20751352/suppress-warning-messages-using-mysql-from-within-terminal-but-password-written)
2. We now insert N records using 1 insert statement instead of N statements.
3. Always include `SET NAMES binary;` into the `*.sql` dump, so MySQL 5.7's CLI can accept the naughty strings 🤷‍.


GregoryIan:
LGTM

PR #47, opened 2020-03-11
ignore generated column

### LINKED ISSUE #44: Omit generated columns when generating sqls
When table has GENERATED column current dumpling will also generate sqls with the generated values, which will cause an error in downstream dbs.
We need to omit GENERATED columns.

**example**

table-structure:
```sql
CREATE TABLE `t1` (
  `id` int(11) DEFAULT NULL,
  `name` varchar(20) DEFAULT NULL,
  `info` json DEFAULT NULL,
  `gen_id` int(11) GENERATED ALWAYS AS (json_extract(`info`,'$.id')) VIRTUAL,
  `info2` varchar(40) DEFAULT NULL,
  UNIQUE KEY `gen_idx` (`gen_id`),
  KEY `multi_col` (`id`,`gen_id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```
sqls generated by **dumpling**:
```sql
/*!40101 SET NAMES binary*/;
INSERT INTO `t1` VALUES
(1, 'Arya Stark', '{"id": 120}', 120, NULL),
(2, 'Catelyn Stark', '{"id": 121}', 121, NULL),
(3, 'Eddard Stark', '{"id": 122}', 122, NULL);
```
sqls generated by **mydumper**:
```sql
/*!40101 SET NAMES binary*/;
/*!40014 SET FOREIGN_KEY_CHECKS=0*/;
INSERT INTO `t1` (`id`,`name`,`info`,`info2`) VALUES
(1,'Arya Stark',CONVERT('{\"id\": 120}' USING UTF8MB4),NULL),
(2,'Catelyn Stark',CONVERT('{\"id\": 121}' USING UTF8MB4),NULL),
(3,'Eddard Stark',CONVERT('{\"id\": 122}' USING UTF8MB4),NULL);
```

3pointer:
resolve #44

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/pingcap/log v0.0.0-20200117041106-d28c14d3b1cd
 	github.com/pingcap/tidb-tools v4.0.0-beta.1.0.20200306103835-530c669f7112+incompatible
 	github.com/pkg/errors v0.9.1
+	github.com/soheilhy/cmux v0.1.4
 	github.com/spf13/cobra v0.0.6
 	go.uber.org/zap v1.14.0
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/pingcap/log v0.0.0-20200117041106-d28c14d3b1cd
 	github.com/pingcap/tidb-tools v4.0.0-beta.1.0.20200306103835-530c669f7112+incompatible
 	github.com/pkg/errors v0.9.1
-	github.com/soheilhy/cmux v0.1.4
 	github.com/spf13/cobra v0.0.6
 	go.uber.org/zap v1.14.0
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -283,39 +283,34 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
 	return withStack(err)
 }
 
-func existsGeneratedFields(db *sql.DB, dbName, tableName string) (bool, error) {
-	query := `SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA=?
-		AND TABLE_NAME=? AND EXTRA REGEXP '(STORED|VIRTUAL) GENERATED';`
-	row := db.QueryRow(query, dbName, tableName)
-	detected := 0
-	err := row.Scan(&detected)
-	if err != nil {
-		if err == sql.ErrNoRows {
-			return false, nil
-		} else {
-			return false, withStack(errors.WithMessage(err, query))
-		}
-	}
-	return detected == 1, nil
-}
-func buildInsertableFields(db *sql.DB, dbName, tableName string) (string, error) {
-	query := `SELECT COLUMN_NAME FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA=?
-		AND TABLE_NAME=? and EXTRA NOT REGEXP '(STORED|VIRTUAL) GENERATED';`
+func buildSelectField(db *sql.DB, dbName, tableName string) (string, error) {
+	query := `SELECT COLUMN_NAME,EXTRA FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA=? AND TABLE_NAME=?;`
 	rows, err := db.Query(query, dbName, tableName)
 	if err != nil {
 		return "", err
 	}
 	defer rows.Close()
 	availableFields := make([]string, 0)
+
+	hasGenerateColumn := false
 	var fieldName string
+	var extra string
 	for rows.Next() {
-		err = rows.Scan(&fieldName)
+		err = rows.Scan(&fieldName, &extra)
 		if err != nil {
 			return "", withStack(errors.WithMessage(err, query))
 		}
-		availableFields = append(availableFields, fieldName)
+		switch extra {
+		case "STORED GENERATED", "VIRTUAL GENERATED":
+			hasGenerateColumn = true
+			continue
+		}
+		availableFields = append(availableFields, wrapBackTicks(fieldName))
 	}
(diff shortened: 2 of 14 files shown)


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/pingcap/log v0.0.0-20200117041106-d28c14d3b1cd
 	github.com/pingcap/tidb-tools v4.0.0-beta.1.0.20200306103835-530c669f7112+incompatible
 	github.com/pkg/errors v0.9.1
+	github.com/soheilhy/cmux v0.1.4
```
kennytm: why?

v4/export/sql.go
```diff
@@ -275,6 +282,41 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
 	return withStack(err)
 }
 
+func existsGeneratedFields(db *sql.DB, dbName, tableName string) (bool, error) {
```
kennytm: This function seems unnecessary. `buildSelectField` can call `buildInsertableFields` directly. The side effect skipping this check is we'll always run `SELECT all, the, columns` rather than `SELECT *`, but I don't think that's a bad thing.

v4/export/sql.go
```diff
@@ -186,9 +193,9 @@ func SelectTiDBRowID(db *sql.DB, database, table string) (bool, error) {
 	return true, nil
 }
 
-func GetColumnTypes(db *sql.DB, database, table string) ([]*sql.ColumnType, error) {
-	query := fmt.Sprintf("SELECT * FROM %s.%s LIMIT 1", database, table)
-	rows, err := db.Query(fmt.Sprintf("SELECT * FROM %s.%s LIMIT 1", database, table))
+func GetColumnTypes(db *sql.DB, field, database, table string) ([]*sql.ColumnType, error) {
```
GregoryIan:
```suggestion
func GetColumnTypes(db *sql.DB, fields, database, table string) ([]*sql.ColumnType, error) {
```

v4/export/sql.go
```diff
@@ -275,6 +282,41 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
...
+	err := row.Scan(&detected)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return false, nil
+		} else {
+			return false, withStack(errors.WithMessage(err, query))
+		}
+	}
+	return detected == 1, nil
+}
```
GregoryIan:
```suggestion
}

```

v4/export/sql.go
```diff
@@ -275,6 +282,41 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
...
+func existsGeneratedFields(db *sql.DB, dbName, tableName string) (bool, error) {
+	query := `SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA=?
+		AND TABLE_NAME=? AND EXTRA REGEXP '(STORED|VIRTUAL) GENERATED';`
+	row := db.QueryRow(query, dbName, tableName)
+	detected := 0
+	err := row.Scan(&detected)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return false, nil
+		} else {
```
GregoryIan:
```suggestion
		}
```

v4/export/sql.go
```diff
@@ -275,6 +282,41 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
...
+			return false, withStack(errors.WithMessage(err, query))
+		}
```
GregoryIan:
```suggestion
```

v4/export/sql.go
```diff
@@ -275,6 +282,41 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
...
+	query := `SELECT 1 FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_SCHEMA=?
+		AND TABLE_NAME=? AND EXTRA REGEXP '(STORED|VIRTUAL) GENERATED';`
+	row := db.QueryRow(query, dbName, tableName)
+	detected := 0
+	err := row.Scan(&detected)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return false, nil
+		} else {
+			return false, withStack(errors.WithMessage(err, query))
```
GregoryIan:
```suggestion
		return false, withStack(errors.WithMessage(err, query))
```

v4/export/sql.go
```diff
@@ -275,6 +282,41 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
...
+	}
+	defer rows.Close()
+	availableFields := make([]string, 0)
+	var fieldName string
+	for rows.Next() {
+		err = rows.Scan(&fieldName)
+		if err != nil {
+			return "", withStack(errors.WithMessage(err, query))
+		}
+		availableFields = append(availableFields, fieldName)
```
GregoryIan: should we quote the fileName using ``

go.mod
GregoryIan: +1

v4/export/sql.go
```diff
@@ -126,9 +131,11 @@ func SelectAllFromTable(conf *Config, db *sql.DB, database, table string) (Table
 	}, nil
 }
 
-func buildSelectQuery(database, table string, where string, orderByClause string) string {
+func buildSelectQuery(database, table string, field string, where string, orderByClause string) string {
```
GregoryIan:
```suggestion
func buildSelectQuery(database, table string, fields string, where string, orderByClause string) string {
```

v4/export/sql.go
GregoryIan: I also think we don't need it

go.mod
3pointer: after I ran`make test`, this go mod will be auto generated, should I remove it?

v4/export/sql.go
3pointer: sure, put logic into [`buildSelectField`](https://github.com/pingcap/dumpling/pull/47/files#diff-5c2ce07cc844075b8e4dfe29c8e256f1R286), use one query to achieve that

go.mod
lichunzhu: No, this is added because I forget to add cmux in last PR.

GregoryIan:
LGTM

PR #49, opened 2020-03-13
Add enable backslash

lichunzhu:
Support using backslash to escape in dumpling.
example from mydumper:
```sql
INSERT INTO `t2` VALUES
(3,200003,'MWQeWw\"\"\'rNmtGxzGp');
```


cmd/dumpling/main.go
```diff
@@ -77,6 +79,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&statusAddr, "status-addr", ":8281", "dumpling API server and pprof addr")
 	rootCmd.PersistentFlags().Uint64VarP(&rows, "rows", "r", export.UnspecifiedSize, "Split table into chunks of this many rows, default unlimited")
 	rootCmd.PersistentFlags().StringVar(&where, "where", "", "Dump only selected records")
+	rootCmd.PersistentFlags().BoolVar(&escapeBackslash, "escape-backslash", false, "use backslash to escape quotation marks")
```
kennytm: The default should be `true` and has the same meaning of (inverse of) [`NO_BACKSLASH_ESCAPES`](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_no_backslash_escapes).

If `escape-backslash=true`, replace `'` by `\'` and `\` by `\\` (and handle `\0`, `\n`, etc. if needed)

If `escape-backslash=false`, replace `'` by `''` and don't touch anything else.

lichunzhu: Revised, PTAL

lichunzhu:
@kennytm PTAL

v4/export/sql_type.go
```diff
@@ -121,17 +223,12 @@ func (s *SQLTypeString) ReportSize() uint64 {
 }
 func (s *SQLTypeString) ToString() string {
 	if s.Valid {
-		return fmt.Sprintf(`'%s'`, escape(s.String))
+		return fmt.Sprintf(`'%s'`, globalEscape.Escape(s.String))
```
GregoryIan: why not return `globalEscape.Escape(s.String)` directly?

v4/export/dump.go
```diff
@@ -13,6 +13,9 @@ import (
 )
 
 func Dump(conf *Config) (err error) {
+	if !conf.EscapeBackslash {
+		globalEscape = &noBackslashEscape{}
```
GregoryIan: put it into config,init or config.adjust? same with L31

v4/export/sql_type.go
```diff
@@ -41,6 +42,107 @@ var dataTypeBin = []string{
...
+			escape = '\\'
+			break
+		case '\'':
+			escape = '\''
+			break
+		}
+
+		if escape != 0 {
+			if last == 0 {
+				bf.Grow(2 * len(s))
```
GregoryIan: why grow 2 times?

GregoryIan:
Rest LGTM

v4/export/sql_type.go
lichunzhu: because we need to use quotation marks ('') to annotate string type sql variable.

v4/export/dump.go
lichunzhu: addressed in https://github.com/pingcap/dumpling/pull/49/commits/e920a47b8342631c4838ef05b2185572e7d754dd

v4/export/prepare.go
```diff
@@ -8,6 +8,10 @@ import (
 )
 
 func adjustConfig(conf *Config) {
+	if !conf.EscapeBackslash {
+		globalEscape = &noBackslashEscape{}
+	}
```
kennytm: When I said "global config" I mean a config variable used by every procedure, not a literal global variable.

v4/export/sql_type.go
```diff
@@ -41,6 +42,107 @@ var dataTypeBin = []string{
...
+		escape byte
+		last   = 0
+	)
+	for i := 0; i < len(s); i++ {
+		escape = 0
+
+		// `'` -> `''` and `\` -> `\\`
+		switch s[i] {
+		case '\\':
+			escape = '\\'
```
kennytm: No need to escape \ when NO_BACKSLASH_ESCAPE. \ is not special.

v4/export/config.go
```diff
@@ -61,6 +62,12 @@ func (conf *Config) getDSN(db string) string {
 	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?charset=utf8mb4", conf.User, conf.Password, conf.Host, conf.Port, db)
 }
 
+type GlobalConfig struct {
+	EscapeBackslash bool
+}
+
+var globalConfig GlobalConfig
```
kennytm: No I mean passing the config into `ToString()`. Are you supposed to rely on a frequently changing global variable in DM?

lichunzhu: Do you mean to pass the config from `v4/export/dump.go:Dump` -> `dumpDatabases` -> ... -> `ToString()`?

kennytm: It should be passed into in the SQL implementation of `WriteInsert`, because whether to escape backslash is an SQL-output-only concern. (But for now the abstraction is totally missing :shrug:).

lichunzhu: Revised in https://github.com/pingcap/dumpling/pull/49/commits/4b1c80c39524bf501a5dcc7d3893f70109592500, PTAL

PR #53, opened 2020-03-18
optimize dumpling's effiency hotfix

lichunzhu:
`bytes.Buffer.Reset` might make `Buffer.Bytes` not valid. So we get a new `bytes.Buffer` from `sync.Pool` every time we want to write buffer to `os.File`. The older `bytes.Buffer` will be put back to `sync.Pool` and reused after its bytes has been written by `os.File`.

Running time:
```
real	0m27.104s
user	0m24.839s
sys	0m8.317s
```


lichunzhu:
@kennytm PTAL

v4/export/writer_util.go
```diff
@@ -148,8 +152,11 @@ func WriteInsert(tblIR TableDataIR, w io.Writer) error {
 			counter += 1
 
 			if bf.Len() >= lengthLimit {
-				wp.input <- bf.Bytes()
-				bf.Reset()
+				wp.input <- bf
+				bf = pool.Get().(*bytes.Buffer)
+				if bf.Cap() < lengthLimit {
+					bf.Grow(lengthLimit)
```
kennytm: Sounds like this should be `lengthLimit - bf.Cap()` 🤔

lichunzhu: Yes, It makes sense.

PR #54, opened 2020-04-15
log: rename all log.Zap to log

AndrewDi:
log.go wrap command logger level, rename all log.Zap to call log wrap method

### DIFF
--- a/v4/log/log.go
+++ b/v4/log/log.go
@@ -49,8 +49,8 @@ func InitAppLogger(cfg *Config) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	appLogger = Logger{logger}
 	logger = appLogger.WithOptions(zap.AddCallerSkip(1))
+	appLogger = Logger{logger}
 	appLevel = props.Level
 	return nil
 }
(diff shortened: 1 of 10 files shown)


v4/log/log.go
```diff
@@ -50,6 +50,7 @@ func InitAppLogger(cfg *Config) error {
 		return errors.WithStack(err)
 	}
 	appLogger = Logger{logger}
+	logger = appLogger.WithOptions(zap.AddCallerSkip(1))
```
kennytm, line 53: I don't think this has any effect after `appLogger` has been assigned. This line should be placed above line 52.

AndrewDi: ok

v4/log/log.go
```diff
@@ -49,6 +49,7 @@ func InitAppLogger(cfg *Config) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
+	logger = appLogger.WithOptions(zap.AddCallerSkip(1))
```
kennytm, line 52:
```suggestion
	logger = logger.WithOptions(zap.AddCallerSkip(1))
```

PR #56, opened 2020-04-17
consistency: enhance auto resolve consistency

AndrewDi:
When use tidb without tikv, dumpling should not set snapshot consistency.

### DIFF
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -293,10 +293,7 @@ func CheckTiDBWithTiKV(db *sql.DB) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	if count > 0 {
-		return true, nil
-	}
-	return false, nil
+	return count > 0, nil
 }
 
 func buildSelectField(db *sql.DB, dbName, tableName string) (string, error) {
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -294,7 +294,7 @@ func CheckTiDBWithTiKV(db *sql.DB) (bool, error) {
 		return false, err
 	}
 	if count > 0 {
-		return true, err
+		return true, nil
 	}
 	return false, nil
 }
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -289,7 +289,7 @@ func CheckTiDBWithTiKV(db *sql.DB) (bool, error) {
 	handleOneRow := func(rows *sql.Rows) error {
 		return rows.Scan(&count)
 	}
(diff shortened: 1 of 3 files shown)


v4/export/sql.go
```diff
@@ -284,6 +284,21 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
...
+	var count int
+	handleOneRow := func(rows *sql.Rows) error {
+		return rows.Scan(&count)
+	}
+	err := simpleQuery(db, "SELECT COUNT(1) as c FROM MYSQL.TiDB WHERE VARIABLE_NAME='tikv_gc_enable'", handleOneRow)
+	if err != nil {
+		return false, err
+	}
+	if count > 0 {
+		return true, err
```
lichunzhu, line 297:
```suggestion
		return true, nil
```

AndrewDi: done

v4/export/sql.go
```diff
@@ -284,6 +284,21 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
...
+		return rows.Scan(&count)
+	}
+	err := simpleQuery(db, "SELECT COUNT(1) as c FROM MYSQL.TiDB WHERE VARIABLE_NAME='tikv_gc_enable'", handleOneRow)
+	if err != nil {
+		return false, err
+	}
+	if count > 0 {
+		return true, nil
+	}
+	return false, nil
```
kennytm, line 299:
```suggestion
	return count > 0, nil
```

v4/export/sql.go
```diff
@@ -284,6 +284,21 @@ func SetTiDBSnapshot(db *sql.DB, snapshot string) error {
 	return withStack(err)
 }
 
+func CheckTiDBWithTiKV(db *sql.DB) (bool, error) {
+	var count int
+	handleOneRow := func(rows *sql.Rows) error {
+		return rows.Scan(&count)
+	}
+	err := simpleQuery(db, "SELECT COUNT(1) as c FROM MYSQL.TiDB WHERE VARIABLE_NAME='tikv_gc_enable'", handleOneRow)
```
kennytm, line 292: perhaps better use `tikv_gc_safe_point` here, since that's what the error message talks about.

```suggestion
	err := simpleQuery(db, "SELECT COUNT(1) as c FROM MYSQL.TiDB WHERE VARIABLE_NAME='tikv_gc_safe_point'", handleOneRow)
```

PR #57, opened 2020-04-17
main: add noData\noSchemas cli parameters

AndrewDi:
if noData is specific, dumpling will not dump table data,
if noSchemas is specific, dumpling will not dump table schema.

### DIFF
--- a/v4/export/dump.go
+++ b/v4/export/dump.go
@@ -125,22 +125,30 @@ func dumpDatabases(ctx context.Context, conf *Config, db *sql.DB, writer Writer)
 }
 
 func dumpTable(ctx context.Context, conf *Config, db *sql.DB, dbName string, table *TableInfo, writer Writer) error {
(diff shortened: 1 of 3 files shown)


v4/export/dump.go
```diff
@@ -125,22 +125,29 @@ func dumpDatabases(ctx context.Context, conf *Config, db *sql.DB, writer Writer)
...
-	if err != nil {
-		return err
+	if !conf.NoSchemas {
+		createTableSQL, err := ShowCreateTable(db, dbName, tableName)
+		if err != nil {
+			return err
+		}
+		if err := writer.WriteTableMeta(ctx, dbName, tableName, createTableSQL); err != nil {
+			return err
+		}
```
lichunzhu, line 146: Can we unify these two `if` into one?

AndrewDi: done

PR #62, opened 2020-05-07
dump: add dump data with sql

AndrewDi:
Add cmd parameter, if specific `sql`, dumpling should executer sql and dump result(Only support dump to csv format now)

### DIFF
--- a/cmd/dumpling/main.go
+++ b/cmd/dumpling/main.go
@@ -46,6 +46,7 @@ var (
 	noSchemas     bool
 	noData        bool
 	csvNullValue  string
+	sql           string
 
 	escapeBackslash bool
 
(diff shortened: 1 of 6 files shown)


AndrewDi:
@kennytm PTAL

cmd/dumpling/main.go
```diff
@@ -88,6 +89,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVarP(&noSchemas, "no-schemas", "m", false, "Do not dump table schemas with the data")
 	rootCmd.PersistentFlags().BoolVarP(&noData, "no-data", "d", false, "Do not dump table data")
 	rootCmd.PersistentFlags().StringVar(&csvNullValue, "csv-null-value", "\\N", "The null value used when export to csv")
+	rootCmd.PersistentFlags().StringVarP(&sql, "sql", "s", "", "Dump data with given sql")
```
kennytm, line 92: I'm not so sure about reserving `-s` here, since mydumper uses `-s` for `--statement-size` (which we are using `-S` for that (which in mydumper and mysqldump means `--socket`)).

We can't use `-e`, it means `--extended-insert` in mysqldump and `--build-empty-files` in mydumper.

cc @3pointer

kennytm: Let's decide this in a new PR.

GregoryIan: only leave `--sql` option?

AndrewDi: good idea...

PR #63, opened 2020-05-08
cli: remove cli parameter and add some log

AndrewDi:
remove cli parameter and add some log


AndrewDi:
@kennytm PTAL

AndrewDi:
it seems like cli flag already remove，close this pr

PR #67, opened 2020-05-11
Delete fix.sql

GregoryIan:


kennytm:
could you also add a `.gitignore`? this file will be generated by `sync-diff-inspector` used in several tests. thanks.

Issue #95, opened 2020-06-08
Let Dumpling set GC safepoint ttl with new pd interface


lichunzhu:
When Dumpling dumps large data it needs to set TiDB snapshot at a ts. When the dumping time is too long tikv may gc this TiDB snapshot ts and Dumpling will fail to dump, exits with an error.

To avoid this problem, users need to set `tikv_gc_life_time` longer, which is explained in the end of https://pingcap.com/docs-cn/stable/export-or-backup-using-dumpling. However, it's not convenient. The users need to run
```sql
update mysql.tidb set VARIABLE_VALUE = '720h' where VARIABLE_NAME = 'tikv_gc_life_time';
```
before dumping and run
```sql
update mysql.tidb set VARIABLE_VALUE = '10m' where VARIABLE_NAME = 'tikv_gc_life_time';
```
after dumping.

But, Dumpling shouldn't change this variable because dumping process may fail in the half.

In v4.0.0, PD has a new interface [UpdateServiceGCSafePoint](https://github.com/pingcap/pd/blob/master/client/client.go#L82) which allows service to set TiKV GC safe point with a ttl, which is very suitable for dumpling's scene. What's more, TiCDC has already used this interface [in their service](https://github.com/pingcap/ticdc/blob/514a16d6b76b08b3a5b7ddb8cc853e1065852246/cdc/owner.go#L418).

So, in summary, I think dumpling can also use this feature to make sure dumped ts is not "GCed".

kennytm:
(I'm setting this P3 because TiDB backup should primarily use BR rather than Dumpling.)

lichunzhu:
> (I'm setting this P3 because TiDB backup should primarily use BR rather than Dumpling.)

Okay

kennytm:
Concern: Dumpling and the TiDB cluster may be in different intranets. When TiDB tells us that PD is on 10.0.0.3:2379, Dumpling may be connecting to a totally irrelevant PD.

![Dumpling connecting to TiDB through a reverse proxy](https://user-images.githubusercontent.com/103023/84111668-02dc6c00-aa5a-11ea-8465-3fe1b963c924.png)

lichunzhu:
> Concern: Dumpling and the TiDB cluster may be in different intranets. When TiDB tells us that PD is on 10.0.0.3:2379, Dumpling may be connecting to a totally irrelevant PD.
> 
[...]

Issue #52, opened 2020-03-18
Add context in `export.Dump`


lichunzhu:
As a lib, dumpling should add `context.Context` in `export.Dump` to shutdown the dump process by using `context.CancelFunc`.

PR #97, opened 2020-06-09
add set gc ts feature

lichunzhu:
Resolve https://github.com/pingcap/dumpling/issues/52 and resolve https://github.com/pingcap/dumpling/issues/95


v4/export/dump.go
```diff
@@ -215,3 +254,31 @@ Loop:
...
+				if err == nil {
+					break
+				}
+				log.Warn("update PD safePoint failed", zap.Error(err))
+				select {
+				case <-ctx.Done():
+					return
+				case <-time.After(time.Second):
+				}
+			}
```
kennytm, line 281: Note that PD and Dumpling are not necessarily living in the same cluster. So failure to update the TTL should *not* leading to this never-ending warning.

v4/export/dump.go
```diff
@@ -215,3 +260,31 @@ Loop:
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-tick.C:
+			log.Debug("update PD safePoint limit with ttl",
+				zap.Uint64("safePoint", snapshotTS),
+				zap.Int64("ttl", ttl))
+			for {
```
kennytm, line 276: also put a retry limit

v4/export/sql.go
```diff
@@ -294,6 +294,39 @@ func ShowMasterStatus(db *sql.DB, fieldNum int) ([]string, error) {
 	return oneRow, nil
 }
 
+func GetPdAddrs(db *sql.DB) ([]string, error) {
+	var pdAddrs []string
+	query := "SELECT * FROM information_schema.cluster_info where type = \"pd\";"
```
kennytm, line 299:
```suggestion
	query := "SELECT * FROM information_schema.cluster_info where type = 'pd';"
```

PR #103, opened 2020-06-18
Refine dumpling user experience

lichunzhu:
1. add dump-empty-database option 
2. check whether database exists before dump
3. allow filetype sql for --Sql command


cmd/dumpling/main.go
```diff
@@ -104,6 +106,8 @@ func main() {
 	pflag.StringVarP(&sql, "sql", "S", "", "Dump data with given sql. This argument doesn't support concurrent dump")
 	pflag.StringArrayVarP(&filters, "filter", "f", []string{"*.*"}, "filter to select which tables to dump")
 	pflag.BoolVar(&caseSensitive, "case-sensitive", false, "whether the filter should be case-sensitive")
+	pflag.BoolVar(&dumpEmptyDatabase, "dump-empty-database", true, "whether to dump empty database")
+	pflag.Uint64Var(&tidbMemQuotaQuery, "tidb-mem-quota-query", export.DefaultTiDBMemQuotaQuery, "The maximum memory available for a single SQL statement. Default: 32GB")
```
kennytm, line 110:
```suggestion
	pflag.Uint64Var(&tidbMemQuotaQuery, "tidb-mem-quota-query", export.DefaultTiDBMemQuotaQuery, "The maximum memory limit for a single SQL statement, in bytes. Default: 32GB")
```

v4/export/dump.go
```diff
@@ -36,6 +36,13 @@ func Dump(conf *Config) (err error) {
 	if err != nil {
 		return err
 	}
+	if conf.ServerInfo.ServerType == ServerTypeTiDB && conf.ServerInfo.ServerVersion.Compare(*tidbV312) >= 0 {
+		log.Info("set tidb_mem_quota_query", zap.Uint64("tidb_mem_quota_query", conf.TiDBMemQuotaQuery))
+		_, err = pool.Exec("set tidb_mem_quota_query = ?", conf.TiDBMemQuotaQuery)
```
kennytm, line 41: should this be moved into `getDSN`?

lichunzhu: sure

lichunzhu: wait a minute, we must make connected db is tidb and then we can set this session variable. how to do this?

kennytm: 🤔 In Lightning if we get the error we'll delete the variable and reconnect. You could use https://pkg.go.dev/github.com/go-sql-driver/mysql?tab=doc#Config.FormatDSN to construct the DSN.

Alternatively, execute this statement every time we perform a `SELECT *`? 

I suspect https://github.com/pingcap/dumpling/issues/104#issuecomment-646445174 is because this `SET` is run on only one connection rather than every connection.

lichunzhu: This is caused because `_tidb_rowid` is distributed sparsely.
image

kennytm: I don't think that's relevant to the OOM

lichunzhu: I tried and it will not return error

lichunzhu: > I don't think that's relevant to the OOM

Why? TiDB gets oom because it can't handle a wide ranged select sql

kennytm:
@cyliu0 can the current PR fix https://github.com/pingcap/dumpling/issues/104#issuecomment-646445174?

README.md
```diff
@@ -20,7 +20,7 @@ Features
 - [x] Export multiple tables in parallel to speed up execution.
 - [ ] Multiple output formats: SQL, CSV, ...
 - [ ] Write to cloud storage (S3, GCS) natively
-- [ ] Advanced table filtering via black-/white-lists
+- [ ] Advanced table filtering via block-/allow-lists
```
kennytm, line 23:
```suggestion
- [x] Advanced table filtering via table filter
```

cyliu0:
> @cyliu0 can the current PR fix [#104 (comment)](https://github.com/pingcap/dumpling/issues/104#issuecomment-646445174)?

Unfortunately not. The current chunk splitting algorithm is not working properly with massive data on TiDB. You have to specify a really small rows argument to workaround.

PR #110, opened 2020-06-29
reduce select usage when query tables

### LINKED ISSUE #109: dumpling costs too much time on getting schemas' tables
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.

Run dumpling with `-T` args to dump only one table. The TiDB cluster has more than 1000 schemas and 100000 tables. Each `select` query will cost about 3s.

2. What did you expect to see?

Dumpling data quickly and successfully.

3. What did you see instead?

Dumpling is blocked at [listAllTables](https://github.com/pingcap/dumpling/blob/master/v4/export/prepare.go#L78).

lichunzhu:
Fix #109 

Select all tables/views at one time.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity
 
### Release note

- Fix the problem that dumpling get blocked when tidb cluster has a lot of tables.

### DIFF
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -77,7 +77,7 @@ func ListAllTables(db *sql.DB, database string) ([]string, error) {
 	return tables.data, nil
 }
 
-func ListAllDatabasesTables(db *sql.DB, allowMap map[string]interface{}, tableType TableType) (DatabaseTables, error) {
+func ListAllDatabasesTables(db *sql.DB, databaseNames []string, tableType TableType) (DatabaseTables, error) {
 	var tableTypeStr string
 	switch tableType {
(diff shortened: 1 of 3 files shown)


v4/export/sql.go
```diff
@@ -77,6 +77,40 @@ func ListAllTables(db *sql.DB, database string) ([]string, error) {
 	return tables.data, nil
```
lichunzhu, line 77: Should we delete `ListAllTables` and `ListAllViews`?

kennytm: we could if nothing uses them. doesn't really matter.

PR #117, opened 2020-07-08
fix concurrency problem

lichunzhu:
1. Now we get tables before we flush all tables, it's not safe when there are tables concurrently written.
2. Support record `show slave status` info in metadata for mysql/mariaDB.
3. Fix the problem that can't set snapshot `lock` when server type is `tidb`.

1. For consistency `flush`: lock tables -> record metadata -> get tables list
    For consistency `lock`:  record metadata -> get tables list -> lock tables
2. Add metadata support in `metadata.go`.
3. Check whether consistency is `snapshot` before add `snapshot` session variable.

### Check List 

Tests 

 - Unit test
 - Integration test
    Manually check metadata for tidb/mysql. Looks fine to me.

Side effects

 - Increased code complexity
 
### Release note

- Fix the bug that flush tables may not work for tables to dump

### DIFF
--- a/v4/export/metadata.go
+++ b/v4/export/metadata.go
@@ -1,20 +1,19 @@
 package export
 
 import (
+	"bytes"
 	"database/sql"
 	"errors"
+	"fmt"
 	"path"
+	"strings"
 	"time"
 )
 
 type globalMetadata struct {
-	logFile string
-	pos     string
-	gtidSet string
+	buffer bytes.Buffer
 
-	filePath   string
-	startTime  time.Time
-	finishTime time.Time
+	filePath string
 }
 
 const (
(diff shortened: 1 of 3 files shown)


v4/export/metadata.go
```diff
@@ -1,20 +1,19 @@
...
-	logFile string
-	pos     string
-	gtidSet string
+	buffer bytes.Buffer
 
-	filePath   string
-	startTime  time.Time
-	finishTime time.Time
+	filePath string
 }
```
kennytm, line 17: why change this to immediately write to a buffer?

lichunzhu: Because I think it's clearer.. WDYT?

kennytm: Heh ok.

v4/export/metadata.go
```diff
@@ -113,16 +100,81 @@ func (m *globalMetadata) getGlobalMetaData(db *sql.DB, serverType ServerType) er
...
+			case "executed_gtid_set":
+				gtidSet = data[i]
+			}
+		}
+		if len(host) > 0 {
+			m.buffer.WriteString("SHOW SLAVE STATUS:\n")
+			if isms {
+				m.buffer.WriteString("\tConnection name: " + connName + "\n")
+			}
+			m.buffer.WriteString(fmt.Sprintf("\tHost: %s\n\tLog: %s\n\tPos: %s\n\tGTID:%s\n\n", host, logFile, pos, gtidSet))
```
kennytm, line 174:
```suggestion
			fmt.Fprintf(&m.buffer, "\tHost: %s\n\tLog: %s\n\tPos: %s\n\tGTID:%s\n\n", host, logFile, pos, gtidSet)
```

PR #119, opened 2020-07-10
fix unique key null bug

lichunzhu:
Fix the problem that when picked up field is unique index NULL but dumpling skipped the NULL value.

Add nullValueCondition to check for the first file like mydumper. A related integration is also added.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Fix the bug that dumpling omit the NULL value when `--r` is specified.

### DIFF
--- a/tests/null_unique_index/run.sh
+++ b/tests/null_unique_index/run.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+set -eu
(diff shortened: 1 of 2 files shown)


lichunzhu:
> LGTM but IMO it's better to avoid using a nullable unique key at all (since multiple NULL can exist)

Yes. This is used for some special situations.

PR #128, opened 2020-08-06
fix consistency bug

lichunzhu:
1. When dumpling use to `unlock tables;` to unlock the FTWRL lock it actually doesn't take effect. Because dumpling use `sql.DB` to unlock tables but `FTWRL` need to be unlocked by the original connection.
2. Currently metadata is recorded after FTWRL is released which is not proper. Metadata info will not be affected by `consistent snapshot`, so we should record it before FTWRL is released.

1. Use one single connection to use FTWRL.
2. Add consistency integration test.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

### Release note

- Fix the bug that FTWRL lock is not released in time.


v4/export/sql.go
```diff
@@ -295,18 +296,18 @@ func GetUniqueIndexName(db *sql.Conn, database, table string) (string, error) {
 	return colName, nil
 }
 
-func FlushTableWithReadLock(db *sql.DB) error {
-	_, err := db.Exec("FLUSH TABLES WITH READ LOCK")
+func FlushTableWithReadLock(db *sql.Conn) error {
+	_, err := db.ExecContext(context.Background(), "FLUSH TABLES WITH READ LOCK")
```
overvenus, line 300: Please avoiding `context.Background()`.

lichunzhu: addressed, PTAL again

v4/export/consistency.go
```diff
@@ -47,34 +54,33 @@ func (c *ConsistencyNone) TearDown() error {
 }
 
 type ConsistencyFlushTableWithReadLock struct {
+	ctx        context.Context
 	serverType ServerType
-	db         *sql.DB
+	conn       *sql.Conn
 }
 
 func (c *ConsistencyFlushTableWithReadLock) Setup() error {
```
overvenus, line 62: What about 
```suggestion
func (c *ConsistencyFlushTableWithReadLock) Setup(ctx context.Context) error {
```

v4/export/consistency.go
```diff
@@ -47,34 +54,33 @@ func (c *ConsistencyNone) TearDown() error {
...
-		return withStack(errors.New("ConsistencyFlushTableWithReadLock lost database connection"))
-	}
-	return UnlockTables(c.db)
+	defer c.conn.Close()
+	return UnlockTables(c.ctx, c.conn)
 }
 
 type ConsistencyLockDumpingTables struct {
-	db        *sql.DB
+	ctx       context.Context
```
overvenus, line 75: Do not store Contexts inside a struct type.

https://golang.org/pkg/context/#:~:text=Do%20not%20store%20Contexts%20inside,first%20parameter%2C%20typically%20named%20ctx%3A&text=See%20https%3A%2F%2Fblog.golang,a%20server%20that%20uses%20Contexts.

tests/consistency/run.sh
```diff
@@ -0,0 +1,60 @@
...
+# build data on mysql
+run_sql "create database $DB_NAME;"
+run_sql "create table $DB_NAME.$TABLE_NAME (a int(255));"
+
+# insert 100 records
+run_sql "insert into $DB_NAME.$TABLE_NAME values $(seq -s, 100 | sed 's/,*$//g' | sed "s/[0-9]*/('1')/g");"
+
+# dumping with consistency flush
+export DUMPLING_TEST_DATABASE=$DB_NAME
+export GO_FAILPOINTS="github.com/pingcap/dumpling/v4/export/ConsistencyCheck=return(\"5s\")"
```
overvenus, line 21:
```suggestion
export GO_FAILPOINTS="github.com/pingcap/dumpling/v4/export/ConsistencyCheck=1*sleep(5000)"
```

v4/export/dump.go
```diff
@@ -165,6 +162,20 @@ func Dump(pCtx context.Context, conf *Config) (err error) {
...
+
+	failpoint.Inject("ConsistencyCheck", func(val failpoint.Value) {
+		interval, err := time.ParseDuration(val.(string))
+		if err != nil {
+			log.Warn("inject failpoint ConsistencyCheck failed", zap.Reflect("value", val), zap.Error(err))
+		} else {
+			log.Info("start to sleep for failpoint ConsistencyCheck", zap.Duration("sleepTime", interval))
+			time.Sleep(interval)
+		}
+	})
```
overvenus, line 177:
```suggestion
	failpoint.Inject("consistency-check", nil)
```

PR #137, opened 2020-08-20
fix binary value null value problem

lichunzhu:
Dumpling doesn't handle null value for binary type variable.

Handle null value when dumping binary columns.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Handle null value when dumping binary columns.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -12,8 +12,9 @@ wget http://download.pingcap.org/tidb-toolkit-$TAG-linux-amd64.tar.gz -O tools.t
 tar -xzvf tools.tar.gz
 mv tidb-toolkit-$TAG-linux-amd64/bin/* bin/
 
+TIDB_TAG="v4.0.4"
 # download tidb-server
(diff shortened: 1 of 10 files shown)


kennytm (commented):
could you add a test case for 

```sql
create table a(c blob);
insert into a values (x'');
insert into a values (null);
```

rest LGTM

install.sh
```diff
@@ -2,9 +2,7 @@
 
 set -e
 
-# FIXME: change to latest version after lightning fix issue
-# https://github.com/pingcap/tidb-lightning/issues/277
-TAG="v3.1.0-beta.1"
+TAG="nightly"
```
kennytm, line 5: seems can't use `"nightly"` here. what about `"latest"`?

lichunzhu: TiDB branch doesn't have a "nightly" tag. I add a single TiDB_TAG for "v4.0.4" now

PR #138, opened 2020-08-20
export GetDSN in dependency usage

lance6716:
export GetDSN to reuse code, when using dumpling as a dependency

⬆️

### Check List 

Tests 

 - Pass CI

Side effects

Related changes


lichunzhu (approved):
LGTM

PR #150, opened 2020-09-09
fix scan null string error

lichunzhu:
Dumpling use `[]string` to parse metadata variables, which can't handle null variable.
```log
[2020/09/09 15:46:07.651 +08:00] [INFO] [log.go:70] ["get global metadata failed"] [task=coa-sync-prod] [unit=dump] [error="sql: Scan error on column index 43, name \"SQL_Remaining_Delay\": converting NULL to string is unsupported"] [errorVerbose="err = sql: Scan error on column index 43, name \"SQL_Remaining_Delay\": converting NULL to string is unsupported
```

Use `[]sql.NullString` instead of `[]string`.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- fix the problem that dumpling fails to parse metadata when some variables are NULL


v4/export/metadata_test.go
```diff
@@ -115,6 +115,34 @@ func (s *testMetaDataSuite) TestMysqlWithFollowersMetaData(c *C) {
...
+	logFile := "ON.000001"
+	pos := "7502"
+	gtidSet := "6ce40be3-e359-11e9-87e0-36933cb0ca5a:1-29"
+	rows := sqlmock.NewRows([]string{"File", "Position", "Binlog_Do_DB", "Binlog_Ignore_DB", "Executed_Gtid_Set"}).
+		AddRow(logFile, pos, "", "", gtidSet)
+	mock.ExpectQuery("SHOW MASTER STATUS").WillReturnRows(rows)
+	mock.ExpectQuery("SELECT @@default_master_connection").WillReturnError(fmt.Errorf("mock error"))
+	mock.ExpectQuery("SHOW SLAVE STATUS").WillReturnRows(sqlmock.NewRows([]string{"SQL_Remaining_Delay"}).AddRow(nil))
+
+	testFilePath := "/test"
```
lance6716, line 134: maybe `c.MkDir` or something in /tmp

lance6716 (commented):
rest LGTM

v4/export/metadata_test.go
lichunzhu: It won't write file to disk since we don't use `writeGlobalMetaData`. So I think it's okay.

lance6716:
LGTM

kennytm (approved):

PR #155, opened 2020-09-21
Write directly to S3

tirsen:
#8 Support S3

Switch to using [VFS](github.com/c2fo/vfs) which supports local files, S3 and GCS. For now this only supports S3 but it should be easy to add support to e.g. GCS.

You simply prefix the output directory with "s3://bucket" e.g.:
```
dumpling -o s3://mybycket/dir
```

For now all config is passed in using the standard AWS environment variables.

### Check List 

Tests

 - Unit test
 - Integration test
Will test by running dumpling against our staging databases.

Side effects

 - Introduces a new dependency on github.com/c2fo/vfs

Related changes

### Release note

- support writing directly to S3


kennytm:
Hi @tirsen, thanks for the contribution! However, we're more interested in including S3 support through the [`github.com/pingcap/br/pkg/storage`](https://pkg.go.dev/github.com/pingcap/br/pkg/storage) shared by all migration tools, which we can also get GCS support for free. Could you see if it's possible to rewrite this using the ExternalStorage interface?

tirsen:
It's not trivial. Afaict there is no `Writer` implementation which means I would have to implement one that buffers and then sends a chunk at a time using the `Uploader`. All doable of course but, yeah, not trivial. :-)

I could implement a `Writer` on top of `ExternalStorage` or is there one already?

tirsen:
Dumpling already does some level of "buffering" itself for example [here](https://github.com/pingcap/dumpling/blob/c4e4eb0a03f2359c853d46b213c647347d9608e6/v4/export/writer_util.go#L192) but the buffer size is [hard coded to be 1mb](https://github.com/pingcap/dumpling/blob/c4e4eb0a03f2359c853d46b213c647347d9608e6/v4/export/writer_util.go#L18) which is not necessarily optimal for S3. It needs to be configurable. It would be better to implement buffering at the `github.com/pingcap/br/pkg/storage` layer so that the configuration option for chunk size is consistent across clients.

tirsen:
Actually vfs isn't a good choice for this use case anyway, it buffers the entire content of a file in memory and sends all of it on close: https://github.com/C2FO/vfs/blob/master/backend/s3/file.go#L203

Or does dumpling split up tables into many smaller files? How large does individual files become if we dump say a 300gb large database?

kennytm:
@tirsen 

> I could implement a `Writer` on top of `ExternalStorage` or is there one already?

Yes you could implement a `Writer` on top of `ExternalStorage`

> Dumpling already does some level of "buffering" itself for example here but the buffer size is hard coded to be 1mb which is not necessarily optimal for S3. It needs to be configurable. It would be better to implement buffering at the github.com/pingcap/br/pkg/storage layer so that the configuration option for chunk size is consistent across clients.

Yes!

> Or does dumpling split up tables into many smaller files? How large does individual files become if we dump say a 300gb large database?

If you supply `-F` or `-r`, dumpling do split up the table into smaller files according to file size or row count.

tirsen:
@kennytm Here is the Writer implementation on top of Uploader: https://github.com/pingcap/br/pull/524

tirsen:
Ok I think this is good to go now

lichunzhu:
Nice job! We will take a look soon.

v4/export/config.go
```diff
@@ -117,6 +129,53 @@ func (conf *Config) GetDSN(db string) string {
...
+	var err error
+	// TODO support other filesystems
+	options := &config.BackendOptions
+	// used for testing
+	endpoint := os.Getenv("DUMPLING_S3_ENDPOINT")
+	if endpoint != "" {
+		options.S3.Endpoint = endpoint
+		options.S3.ForcePathStyle = true
+		options.GCS.Endpoint = endpoint
+	}
```
lichunzhu, line 155: I think we can use [failpoint](https://github.com/pingcap/failpoint) here like we did in https://github.com/pingcap/dumpling/blob/48b61dc62c7e27c99b81d939e4614f07e3f3a9d2/v4/export/dump.go#L194 and https://github.com/pingcap/dumpling/blob/48b61dc62c7e27c99b81d939e4614f07e3f3a9d2/tests/consistency/run.sh#L21, but this way is also good.

v4/export/config.go
```diff
@@ -62,6 +71,9 @@ type Config struct {
 	SessionParams      map[string]interface{}
 
 	PosAfterConnect bool
+
+	externalStorage     storage.ExternalStorage
```
lichunzhu, line 75: We need to append a `json:"-"` tag to the end of this field so that it won't be printed, just like we did for field `TableFilter`.

v4/export/config.go
```diff
@@ -117,6 +129,53 @@ func (conf *Config) GetDSN(db string) string {
 	return dsn
 }
 
+func (config *Config) ExternalStorage(ctx context.Context) storage.ExternalStorage {
+	config.externalStorageOnce.Do(func() {
```
lichunzhu, line 133: Can we remove `externalStorageOnce` and move the create part to [adjustConfig](https://github.com/pingcap/dumpling/blob/48b61dc62c7e27c99b81d939e4614f07e3f3a9d2/v4/export/prepare.go#L73)?

v4/export/config.go
```diff
@@ -117,6 +129,53 @@ func (conf *Config) GetDSN(db string) string {
...
+		s, err := config.createExternalStorage(ctx)
+		if err != nil {
+			panic(err)
```
lichunzhu, line 136: We'd better return the error to the outside function instead of `panic`. Our users may get confused if they saw panic info.

v4/export/config.go
lichunzhu: @kennytm WDYT?

v4/export/config.go
```diff
@@ -117,6 +129,53 @@ func (conf *Config) GetDSN(db string) string {
...
+	if err != nil {
+		return nil, err
+	}
+	// TODO support sendCreds config option?
+	return storage.Create(ctx, b, true)
+}
+
+func ensureStorageClassPrefix(path string) string {
+	if !regexp.MustCompile("^\\w+://").MatchString(path) {
+		return "file:///" + path
```
lichunzhu, line 166: If I use a relative path here, for example, `./export-2020-09-25T16:48:39+08:00`, this function will return `file:///./export-2020-09-25T16:48:39+08:00` which seems incorrect.

lichunzhu: I will take a look on how to solve this problem.

lichunzhu: We need to change the relative directory to an abs one. I will fire a PR in BR to solve this problem.

lichunzhu: https://github.com/pingcap/br/pull/531

v4/export/config.go
kennytm: +1

v4/export/config.go
tirsen: failpoint looks neat, I'll make it so!

v4/export/config.go
tirsen: Good idea!

v4/export/config.go
tirsen: nice!

tirsen:
Ok I think all comments resolved now. PHAL

lichunzhu (commented):
Can we add a configuration `region` to set region for s3 configurations? When I'm testing upload to S3, setting `AWS_REGION` doesn't work. If `options.S3.Region` is set it will work correctly.
Rest LGTM, @kennytm PTAL

kennytm:
@lichunzhu using the URL `s3://bucket/path?region=us-west-2` should work.

lichunzhu:
> @lichunzhu using the URL `s3://bucket/path?region=us-west-2` should work.

Yes it works. So we just ask users to set `region` in the `OutputDir` URL?

tests/s3/run.sh
```diff
@@ -0,0 +1,53 @@
...
+run_sql "create table $DB_NAME.$TABLE_NAME (a int(255));"
+
+# insert 100 records
+run_sql "insert into $DB_NAME.$TABLE_NAME values $(seq -s, 100 | sed 's/,*$//g' | sed "s/[0-9]*/('1')/g");"
+
+# run dumpling!
+HOST_DIR=${DUMPLING_OUTPUT_DIR}
+export DUMPLING_OUTPUT_DIR=s3://mybucket/dump
+export DUMPLING_TEST_DATABASE=$DB_NAME
+export GO_FAILPOINTS="github.com/pingcap/dumpling/v4/export/ExternalStorageEndpoint=100%return(\"http://localhost:5000\")"
```
kennytm, line 40: this failpoint is unnecessary, you could just set the `DUMPLING_OUTPUT_DIR` to `"s3://bucket/dump?endpoint=http%3A%2F%2Flocalhost%3A5000"` (or you could expose those `--s3.*` options)

v4/export/config.go
```diff
@@ -117,6 +127,28 @@ func (conf *Config) GetDSN(db string) string {
...
+			options.S3.Endpoint = endpoint
+			options.S3.ForcePathStyle = true
+			options.GCS.Endpoint = endpoint
+		}
+	})
+	b, err = storage.ParseBackend(path, options)
+	if err != nil {
+		return nil, err
+	}
+	// TODO support sendCreds config option?
```
kennytm, line 148: no need to support Send Creds since Dumpling works independently.

v4/export/config.go
```diff
@@ -9,12 +10,19 @@ import (
...
+	"github.com/pingcap/kvproto/pkg/backup"
 	filter "github.com/pingcap/tidb-tools/pkg/table-filter"
 	"go.uber.org/zap"
+
+	"github.com/pingcap/dumpling/v4/log"
 )
 
 type Config struct {
+	// TODO support passing in options via flags as well as toml
+	storage.BackendOptions
```
kennytm, line 24: supporting it should be as simple as using [`storage.DefineFlags`](https://pkg.go.dev/github.com/pingcap/br/pkg/storage#DefineFlags) and using [`.ParseFromFlags()`](https://pkg.go.dev/github.com/pingcap/br/pkg/storage#BackendOptions.ParseFromFlags) after constructing the `BackendOptions`.

v4/export/writer_util.go
```diff
@@ -343,65 +347,72 @@ func writeBytes(writer io.Writer, p []byte) error {
...
 		}
-		log.Error("close file failed", zap.String("path", path))
 	}
 	return fileWriter, tearDownRoutine
 }
 
+func encodeFileName(path string) string {
+	return strings.ReplaceAll(path, "%", "%25")
+}
+
```
kennytm, line 415: unused function, delete?

```suggestion
```

v4/export/config.go
tirsen: so we should pass false?

kennytm: any value is fine since it won't be used. i mean there's nothing "TODO" here 😄

tests/s3/run.sh
tirsen: yay!

v4/export/config.go
tirsen: it was indeed simple thanks!

tirsen:
CLI flags are now exposed

kennytm:
> Is `AWS_REGION` a standard env var? 

This env var [should be used by default](https://github.com/aws/aws-sdk-go/blob/38c74caea1398949b67da14dfaa79cabe704a57f/aws/session/env_config.go#L39-L48), but unfortunately we [have overridden](https://github.com/pingcap/br/blob/master/pkg/storage/s3.go#L234) the config making all env var useless.

I think we should fix this later on BR.

v4/export/config.go
```diff
@@ -9,12 +10,17 @@ import (
...
-	"github.com/pingcap/dumpling/v4/log"
+	"github.com/pingcap/br/pkg/storage"
 	filter "github.com/pingcap/tidb-tools/pkg/table-filter"
 	"go.uber.org/zap"
+
+	"github.com/pingcap/dumpling/v4/log"
 )
 
 type Config struct {
+	// TODO support passing in options via flags as well as toml
```
kennytm, line 21:
```suggestion
```

config file (toml support) would be another topic 🙂

kennytm (commented):
rest LGTM, thanks 😄

kennytm (approved):

PR #161, opened 2020-10-02
Use the updated Writer implementation from br

tirsen:


tirsen:
Requires the following to be merged first: https://github.com/pingcap/br/pull/547

kennytm:
as pingcap/br#547 is rejected, should we also close this? @tirsen

tirsen:
Sure!

PR #163, opened 2020-10-11
export: support parallel dump on TiDB clustered index table

tangenta:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -278,6 +278,13 @@ func buildOrderByClause(conf *Config, db *sql.Conn, database, table string) (str
 		return "", nil
 	}
 	if conf.ServerInfo.ServerType == ServerTypeTiDB {
+		if conf.ChunkByTiDBRegion {
+			allPks, err := GetAllPrimaryKeyName(db, database, table)
+			if err != nil {
+				return "", withStack(err)
+			}
+			return fmt.Sprintf("ORDER BY %s", strings.Join(allPks, ", ")), nil
+		}
 		ok, err := SelectTiDBRowID(db, database, table)
 		if err != nil {
 			return "", withStack(err)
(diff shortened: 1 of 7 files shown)


v4/export/sql.go
```diff
@@ -441,6 +442,62 @@ func GetTiDBDDLIDs(db *sql.DB) ([]string, error) {
...
+			log.Warn("unexpected json value type when decoding TiDB region key",
+				zap.String("key", key), zap.String("decode result", decodeResult), zap.Any("handle", v))
+		}
+		ret = append(ret, js)
+	}
+	return ret
+}
+
+func GetTiDBRegionStartKeys(db *sql.Conn, dbName, tableName string) []string {
+	query := "SELECT start_key FROM information_schema.tikv_region_status s, information_schema.tables t " +
```
lichunzhu, line 484: Have you tested this new dumpling manually? For example, primary key whose type is `int`, `varchar`, `datatime`, `date`...

v4/export/sql.go
```diff
@@ -441,6 +442,62 @@ func GetTiDBDDLIDs(db *sql.DB) ([]string, error) {
...
+		"WHERE s.table_id = t.tidb_table_id AND t.db_name = ? AND t.table_name = ? AND is_index = 0 ORDER BY start_key;"
```
lichunzhu, line 485: It sames that `information_schema.tables` doesn't have column `db_name` but have column `table_schema`.
My TiDB version is v4.0.7.

tangenta: This should be `s.db_name`. Fixed.

lichunzhu (commented):
1. It seems that `chunk-by-region` option can't control how much regions dumpling concurrently dump. Should we use `APPROXIMATE_KEYS` or something to control it?
2. Do we need a `chunk-by-region` option? I think we don't need that because I think this split solution is stable enough.
@kennytm WDYT?

tangenta:
Done by https://github.com/pingcap/dumpling/pull/205.

Issue #164, opened 2020-10-12
Support -F even specified -r


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

We can dump a single table concurrently with `-r`, but we may also want to limit the file size with `-F`.

We may support `-r` and `-F` to be specified in one dump round and rotate to a new file when `-r` reached or `-F` reached.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Optimization:**

lichunzhu:
Dumped sql will be labeled as `name.rowIndex.fileIndex.sql` at first. After the dumping process finished, `dumpling` will rename these files to set index from 1 to X.

PR #177, opened 2020-10-27
support both filesize and rows arguments

lichunzhu:
Fix https://github.com/pingcap/dumpling/issues/164

Support configuring both `-F` and `-r` arguments.
When specifying `-F` or `-r` argument, the dumped SQL format is `${db}.${table}.%09d.sql`.
When specifying both `-F` and `-r` arguments, the dumped SQL format is `${db}.${table}.%09d%04d.sql`.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Support configuring both `-F` and `-r` arguments at the same time.


kennytm:
```
tests/rows/run.sh: 57: printf: {1..10}: expected numeric value

Makefile:28: recipe for target 'integration_test' failed
```

v4/export/writer.go
```diff
@@ -137,12 +141,19 @@ type csvOption struct {
...
-	return &outputFileNamer{
-		Index: ir.ChunkIndex(),
+func newOutputFileNamer(ir TableDataIR, bothEnabled bool) *outputFileNamer {
+	o := &outputFileNamer{
 		DB:    ir.DatabaseName(),
 		Table: ir.TableName(),
 	}
+	if bothEnabled {
+		o.id = 0
+		o.idFormat = fmt.Sprintf("%09d", ir.ChunkIndex()) + "%04d"
```
mightyguava, line 151: 13 digits will overflow an int32. Do tools that read these files parse out the number, or just rely on it for filesystem ordering?

v4/export/writer_test.go
```diff
@@ -117,7 +117,7 @@ func (s *testDumpSuite) TestWriteTableData(c *C) {
 	err = writer.WriteTableData(ctx, tableIR)
 	c.Assert(err, IsNil)
 
-	p := path.Join(dir, "test.employee.0.sql")
+	p := path.Join(dir, "test.employee.000000000.sql")
```
mightyguava, line 120: Should there be a test that uses the 13 digit number?

v4/export/writer.go
kennytm: so far all of myloader, Lightning and DM only rely on filesystem order, and won't parse the number. Lightning is the only one requiring that part to be `[0-9]+` because the regex was written too tightly.

v4/export/writer_test.go
lichunzhu: Added in https://github.com/pingcap/dumpling/pull/177/commits/ffcea135e9cbe26c577f2ea7004b3dc89f686044. The integration test in `rows` will also help check this.

v4/export/writer.go
```diff
@@ -126,9 +127,12 @@ func writeMetaToFile(ctx context.Context, target, metaSQL string, s storage.Exte
 type CSVWriter struct{ SimpleWriter }
 
 type outputFileNamer struct {
-	Index int
+	Index string
 	DB    string
 	Table string
+
+	id       int
+	idFormat string
```
kennytm, line 135: can you make the struct like this?

```go
type outputFileNamer struct {
    Index1 int
    Index2 int
    DB     string
    Table  string
    format string
}
```

use `%09[1]d` to format with first index, `%09[2]d` with second index, and `%09d%04d` for both indices

kennytm (dismissed):
LGTM.

Though those Index1 and Index2 could probably use more descriptive names 😉

lichunzhu:
> LGTM.
> 
> Though those Index1 and Index2 could probably use more descriptive names 😉

I revise `Index1` to `ChunkIndex` and `Index2` to `FileIndex`.

kennytm (approved):

kennytm:
(trigger CI)

PR #184, opened 2020-11-03
1. add condition

imiskolee:

the SQL will be output x''  on NULL binary field.

just add a condition for test binary length

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

 - Increased code complexity

Related changes


kennytm:
Thanks for the PR, but evidenced in `tests/e2e/run.sh`, the empty byte string `x''` and `NULL` are two distinct values, so I'm afraid we can't merge this PR like this.

PR #188, opened 2020-11-05
fix tidb master status bug

lichunzhu:
1. For TiDB, metadata will use TSO from `show master status`, which is the current TiDB TSO. If the user sets `--snapshot`, it will be different from that.
2. For TiDB not setting `--snapshot`, we will get a snapshot TSO at first to set PD gc-ttl. After that, we will use `show master status` to get another snapshot TSO, which appears to be different from the first one.

Use snapshot TSO fetched at first expect for consistency lock for TiDB.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- fix the problem that TiDB  pos in metadata is not accurate

### DIFF
--- a/v4/export/dump.go
+++ b/v4/export/dump.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


kennytm (approved):
LGTM

PR #190, opened 2020-11-06
avoid get blocked in dumping when mysql connection is broken

lichunzhu:
### What problem does this PR solve? 
Fix #181 

`dumpling` may get blocked at fetching data from database server when the connection to the database server is closed due to some reason.

Here is the goroutine information:
```
goroutine 265 [IO wait]:
internal/poll.runtime_pollWait(0x8c159d8, 0x72, 0xffffffffffffffff)
	/usr/local/go/src/runtime/netpoll.go:184 +0x55
internal/poll.(*pollDesc).wait(0xc00064c698, 0x72, 0xf00, 0xfc8, 0xffffffffffffffff)
	/usr/local/go/src/internal/poll/fd_poll_runtime.go:87 +0x45
internal/poll.(*pollDesc).waitRead(...)
	/usr/local/go/src/internal/poll/fd_poll_runtime.go:92
internal/poll.(*FD).Read(0xc00064c680, 0xc000820038, 0xfc8, 0xfc8, 0x0, 0x0, 0x0)
	/usr/local/go/src/internal/poll/fd_unix.go:169 +0x22b
net.(*netFD).Read(0xc00064c680, 0xc000820038, 0xfc8, 0xfc8, 0xc000022600, 0xc000897000, 0x64)
	/usr/local/go/src/net/fd_unix.go:202 +0x4f
net.(*conn).Read(0xc000a6e0a0, 0xc000820038, 0xfc8, 0xfc8, 0x0, 0x0, 0x0)
	/usr/local/go/src/net/net.go:184 +0x68
github.com/go-sql-driver/mysql.(*buffer).fill(0xc000213e60, 0xc6, 0x0, 0x0)
	/Users/chauncy/code/goPath/pkg/mod/github.com/go-sql-driver/mysql@v1.5.0/buffer.go:90 +0x13c
github.com/go-sql-driver/mysql.(*buffer).readNext(0xc000213e60, 0xc6, 0xc000820fc4, 0x4, 0x3c, 0x0, 0x0)
	/Users/chauncy/code/goPath/pkg/mod/github.com/go-sql-driver/mysql@v1.5.0/buffer.go:119 +0x9c
github.com/go-sql-driver/mysql.(*mysqlConn).readPacket(0xc000213e60, 0x5668bc1, 0x711b3a7, 0x53, 0xc0, 0x71121e8)
	/Users/chauncy/code/goPath/pkg/mod/github.com/go-sql-driver/mysql@v1.5.0/packets.go:67 +0x120
[...]

PR #194, opened 2020-11-09
filter: refine system database filter rule

lichunzhu:
close https://github.com/pingcap/dumpling/pull/190

excluding system databases for both TiDB and MySQL. Users can reset `--filter` argument to dump system databases.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- Exclude system databases in `--filter` parameter by default.


v4/export/config.go
```diff
@@ -154,6 +154,7 @@ const (
 	DefaultTiDBMemQuotaQuery = 32 * (1 << 30)
 	DefaultStatementSize     = 1000000
 	TiDBMemQuotaQueryName    = "tidb_mem_quota_query"
+	DefaultTableFilter       = "!/^(mysql|test|INFORMATION_SCHEMA|PERFORMANCE_SCHEMA|METRICS_SCHEMA|INSPECTION_SCHEMA)$/.*"
```
kennytm, line 157:
```suggestion
	DefaultTableFilter       = "!/^(mysql|sys|INFORMATION_SCHEMA|PERFORMANCE_SCHEMA|METRICS_SCHEMA|INSPECTION_SCHEMA)$/.*"
```

i see no reason excluding the `test` schema. but [`sys`](https://dev.mysql.com/doc/mysql-monitor/8.0/en/mem-graphs-sys-schema-install-ref.html) *is* a system schema.

(btw why not `[]string{"*.*", "!mysql.*", "!sys.*", "!INFORMATION_SCHEMA.*", "!PERFORMANCE_SCHEMA.*", "!METRICS_SCHEMA.*", "!INSPECTION_SCHEMA.*"}`)

kennytm (approved):

v4/export/config.go
lichunzhu: Yes. I made a mistake. It should be `sys` here.

Because I think that's easier for users to paste if they want to exclude some tables and system tables.
If users want to exclude some databases they can type fewer lines if they have seen the help information. Or, they need to type `--f '*.*' -f '!mysql.*' ... -f '${the tables they want to exclude}'`

Issue #178, opened 2020-10-30
Retry failed chunks


mightyguava:
## Feature Request

**Is your feature request related to a problem? Please describe:**
For really large databases, the likelihood that a database connection is interrupted during a dump increases. Right now, dumpling doesn't retry and just fails the chunk, logging an error message, and then failing the dump after the entire export is finished.

**Describe the feature you'd like:**
It would be nice if dumpling could retry with an exponential backoff. I would also include a flag so that dumpling can exit immediately if all retries for a chunk fail.

Here's the proposed new flags and behavior

`--retry-chunks=N` where N indicates the number of times to retry each chunk
`--retry-initial-backoff=D` where D is a duration indicating how long we wait before the first retry, doubling it each subsequent retry.
`--fail-fast` where if a single chunk fails after all retries, the export is immediately aborted

For us, where we use the [RDS IAM Auth Token](https://aws.amazon.com/premiumsupport/knowledge-center/users-connect-rds-iam/) to authenticate, this poses an additional challenge that the token is only valid for 15 minutes after being created. A retry on a failed connection hours later would not be able to authenticate, so we'd need support in dumpling to fetch a new token rather than having it provided. I'll file a separate ticket for this.

lichunzhu:
I think this makes sense. It seems not hard to accomplish this and it's helpful for dumping a large database. Maybe [dumpling checkpoint](https://github.com/pingcap/dumpling/issues/10) can also solve this problem but it's hard to complete that feature now.
WDYT? @kennytm

kennytm:
I think we could just retry without adding any config parameters (always use the default values)

lichunzhu:
Implemented in https://github.com/pingcap/dumpling/pull/182

lichunzhu:
@mightyguava After this PR https://github.com/pingcap/dumpling/pull/199 you can set `--transactional-consistency` to false so that dumpling can keep the lock during the whole dumping progress and dumpling will try to rebuild MySQL connection when facing an error. Thank you for your advice!

PR #199, opened 2020-11-12
rebuild mysql conn when retry failed chunks and support `trx-consistency-only` parameter

lichunzhu:
https://github.com/pingcap/dumpling/issues/178
When we retry on failed chunks, if mysql connection is broken, the retry will fail directly and will take no effect.

1. Add `trx-consistency-only` option. The default value is true. If it is `false`, dumpling won't release locks until the dumping progress is finished.
2. When dumpling retrys on failed chunks, we should rebuild MySQL connection when it was broken.
* For consistency none and consistency snapshot, we can rebuild MySQL connection directly since it doesn't assure consistent snapshot. But for consistency none we should update metadata after conn for DM when we build new MySQL conn.
* For consistency flush and consistency lock, we can rebuild MySQL connection only when dumpling hold locks during the whole dumping progress (`trx-consistency-only` = false). Before rebuilding MySQL connection we should also check that whether the mysql lock connection is still alive.

### Check List 

Tests 

 - Unit test
 - Integration test
    Test kill the socket mysql connection. The connection can be rebuilt and dump data correctly.

Side effects

 - Increased code complexity

Related changes

### Release note

- Support `trx-consistency-only` parameter and support rebuild mysql connections during retry.

### DIFF
--- a/v4/export/config.go
+++ b/v4/export/config.go
@@ -138,7 +138,7 @@ func DefaultConfig() *Config {
 		SortByPk:           true,
 		Tables:             nil,
 		Snapshot:           "",
-		Consistency:        "auto",
+		Consistency:        consistencyTypeAuto,
 		NoViews:            true,
 		Rows:               UnspecifiedSize,
 		Where:              "",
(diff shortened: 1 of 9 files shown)


lichunzhu:
Should be merged after https://github.com/pingcap/dumpling/pull/190

kennytm:
i suggest we avoid the obscure abbreviation `trx`. 

since `--trx-consistency-only` in mydumper defaults to "false", while here we default to "true", i think we can pick a different name. `--transactional-consistency=0` or `--transactional-snapshot=0` or something.

lichunzhu:
> i suggest we avoid the obscure abbreviation `trx`.
> 
> since `--trx-consistency-only` in mydumper defaults to "false", while here we default to "true", i think we can pick a different name. `--transactional-consistency=0` or `--transactional-snapshot=0` or something.

I think changing the name is okay, but we should keep this behavior the same as before, which means `--transactional-consistency` defaults to `1`.

kennytm:
Yes the default should be 1.

kennytm (dismissed):
LGTM

lichunzhu:
fix some typo in https://github.com/pingcap/dumpling/pull/199/commits/91dcef296b33b9220e28c6c0deea33138e529798

kennytm (approved):

PR #200, opened 2020-11-16
add MySQL chaos test

lichunzhu:
1. `Dumpling` may return error when picking primary integer index because we don't check database name at https://github.com/pingcap/dumpling/blob/master/v4/export/sql.go#L642.
2. `Dumpling` is in lack of chaos test.

1. refine pick mechism
2. add mysql chaos test

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- No release note

### DIFF
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -350,24 +350,17 @@ func GetUniqueIndexName(db *sql.Conn, database, table string) (string, error) {
 }
 
 func getNumericIndex(db *sql.Conn, database, table, indexType string) (string, error) {
-	keyQuery := "SELECT column_name, data_type FROM information_schema.columns " +
-		"WHERE table_schema = ? AND table_name = ? AND column_key = ?;"
-	var colName, dataType string
-	rows, err := db.QueryContext(context.Background(), keyQuery, database, table, indexType)
-	if err != nil {
+	keyQuery := "SELECT column_name FROM information_schema.columns " +
+		"WHERE table_schema = ? AND table_name = ? AND column_key = ? AND data_type IN ('int', 'bigint');"
+	var colName string
+	row := db.QueryRowContext(context.Background(), keyQuery, database, table, indexType)
+	err := row.Scan(&colName)
+	if err == sql.ErrNoRows {
+		return "", nil
+	} else if err != nil {
 		return "", errors.Annotatef(err, "sql: %s, indexType: %s", keyQuery, indexType)
 	}
-	defer rows.Close()
-	for rows.Next() {
-		if err := rows.Scan(&colName, &dataType); err != nil {
-			return "", errors.Annotatef(err, "sql: %s, indexType: %s", keyQuery, indexType)
-		}
-		switch strings.ToLower(dataType) {
-		case "int", "bigint":
-			return colName, nil
-		}
-	}
-	return "", nil
+	return colName, nil
 }
 
 func FlushTableWithReadLock(ctx context.Context, db *sql.Conn) error {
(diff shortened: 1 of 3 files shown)


v4/export/sql.go
```diff
@@ -342,33 +342,32 @@ func GetPrimaryKeyColumns(db *sql.Conn, database, table string) ([]string, error
...
+		return "", errors.Annotatef(err, "sql: %s, indexType: %s", keyQuery, indexType)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		if err := rows.Scan(&colName, &dataType); err != nil {
+			return "", errors.Annotatef(err, "sql: %s, indexType: %s", keyQuery, indexType)
+		}
+		switch strings.ToLower(dataType) {
+		case "int", "bigint":
+			return colName, nil
```
kennytm, line 367: why not move the `dataType` check into the SQL?

```sql
SELECT column_name FROM information_schema.columns
WHERE table_schema = ? AND table_name = ? AND column_key = ? AND data_type IN ('int', 'bigint');
```

v4/export/writer_util.go
```diff
@@ -189,6 +193,19 @@ func WriteInsert(pCtx context.Context, tblIR TableDataIR, w storage.Writer, file
...
 			counter += 1
 			wp.AddFileSize(uint64(bf.Len()-lastBfSize) + 2) // 2 is for ",\n" and ";\n"
+			failpoint.Inject("ChaosBrokenMySQLConn", func(val failpoint.Value) {
+				brokenPoint := val.(int)
+				if counter >= brokenPoint {
+					var err error
+					doOnce.Do(func() {
+						tblIR.Close()
+						err = errors.New("connection is closed")
+					})
```
kennytm, line 203: why use `counter >= brokenPoint` with a `sync.Once` instead of checking `counter == brokenPoint`, or [use failpoint's `prob%count*action` feature](https://github.com/pingcap/failpoint/#key-concepts)

```sh
export GO_FAILPOINTS="github.com/pingcap/dumpling/v4/export/ChaosBrokenMySQLConn=1%1*return(1)" 
```

v4/export/writer_util.go
```diff
@@ -189,6 +190,13 @@ func WriteInsert(pCtx context.Context, tblIR TableDataIR, w storage.Writer, file
 			}
 			counter += 1
 			wp.AddFileSize(uint64(bf.Len()-lastBfSize) + 2) // 2 is for ",\n" and ";\n"
+			failpoint.Inject("ChaosBrokenMySQLConn", func(val failpoint.Value) {
+				brokenPoint := val.(int)
+				if counter >= brokenPoint {
+					tblIR.Close()
+					failpoint.Return(errors.New("connection is closed"))
+				}
```
kennytm, line 198: 🤔 so now we will kill multiple connections?

lichunzhu: No. We will kill connection only for one time.
Setting `ChaosBrokenMySQLConn=$broken_point*return($broken_point)` is to run `brokenPoint` times, but the first `brokenPoint-1` times it won't enter into this if block.

kennytm: so why not use `1%1*return()` which will trigger at 1% probability and run at most 1 time.

lichunzhu: Then maybe sometimes this test can pass and sometimes this test will fail? 🤔

kennytm: You could make the percent higher (e.g. `9%1*return()`) to avoid the "run 0 times" case. Otherwise I don't see this is much different than using `$RANDOM`.

lichunzhu: How about directly failing one time like this?

kennytm: yeah it's fine though it's not really "chaos" now 🙃

kennytm (approved):
LGTM

PR #201, opened 2020-11-17
remove dump_test.go

tangenta:

`dump_test.go` is too hard to maintain because 
- we need to match every query and keep the order
- we need to assign a expected result(which is not easy to write correctly) for every query

So even a tiny change may break this test and it takes a lot of time to figure it out.

The work done by `dump_test.go` is much like "integration test", we should put them to a real integration test.

Remove `dump_test.go` and add an integration test.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

N/A

Related changes

### Release note

No release note


tests/basic/run.sh
```diff
@@ -0,0 +1,32 @@
...
+# Test for simple case.
+run_sql "drop database if exists \`$DB_NAME\`;"
+run_sql "create database \`$DB_NAME\`;"
+run_sql "create table \`$DB_NAME\`.\`$TABLE_NAME\` (a int);"
+run_sql "insert into \`$DB_NAME\`.\`$TABLE_NAME\` values (1), (2);"
+
+run_dumpling
+
+cnt=`grep -w "(.*)" ${DUMPLING_OUTPUT_DIR}/${DB_NAME}.${TABLE_NAME}.000000000.sql|wc -l`
+echo "records count is ${cnt}"
```
lichunzhu, line 18: Should we check the content here?

tangenta: Updated.

lichunzhu:
Should we delete this unit test? @kennytm

kennytm (commented):
lgtm

kennytm (approved):

PR #202, opened 2020-11-20
Add more log and compress option

lichunzhu:
dumpling needs to provide users more progress info and metrics while and after dumping.

1. Add more status log
2. Add `--compress` option
3. Add Prometheus handler

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

### Release note

- support --compress option


v4/export/config.go
```diff
@@ -233,6 +237,7 @@ func (conf *Config) DefineFlags(flags *pflag.FlagSet) {
 	flags.Duration(flagReadTimeout, 15*time.Minute, "I/O read timeout for db connection.")
 	flags.MarkHidden(flagReadTimeout)
 	flags.Bool(flagTransactionalConsistency, true, "Only support transactional consistency")
+	flags.BoolP(flagCompress, "c", false, "Compress output files")
```
kennytm, line 240: is it possible to make this option take "gz" (or "gzip") as input? to allow for other compression formats

kennytm (commented):
Rest LGTM

v4/export/config.go
lichunzhu: addressed in https://github.com/pingcap/dumpling/pull/202/commits/8a5fe5de35a68dcb3dbe0b3620558a49bcf192a3

kennytm (approved):

PR #205, opened 2020-11-27
refactor writer.go and support concurrent dump by tablesample

tangenta:

This PR utilize the `TABLESAMPLE` feature provided by https://github.com/pingcap/tidb/pull/20883 to implement concurrent dumping on a single table.

There are 2 ways to dump tables: `sequentialDumpTable` and `concurrentDumpTable`. For the TiDBs with version >= `5.0.0`(which contain the `TABLESAMPLE` feature), a specialized function `concurrentDumpTiDBTables` is used.

- The `Writer` interface is changed to `Writer` struct because this abstraction is neither extensible nor convienent.
- During dumping, 
  - multiple `Writer`s are created and there is one connection(`*sql.Conn`) per `Writer`.
  - The database/table meta information and table data are packed into a unit `Task` and sent to a global task channel.
  - Each writer pulls the task from channel constantly, and writes them to the storage.
- This PR also refactors some of the code.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test
   1. Run `make integration_test` first to generate testing data.
   1. Compile the TiDB from https://github.com/pingcap/tidb/pull/20883 and replace it into `$dumpling_base_dir/bin/tidb-server`.
   1. Change the `tableSampleVersion` in `config.go` to a value that <= `4.0.0` (like `3.0.0`)
   1. Dump the testing data.
   1. Run `make integration_test`.

A simple manual test for clustered index is also passed.

Side effects

N/A

Related changes

N/A
 
### Release note

- No release note

### DIFF
--- a/v4/export/dump.go
+++ b/v4/export/dump.go
@@ -162,8 +162,8 @@ func (d *Dumper) Dump() (err error) {
 
 	taskChan := make(chan Task, defaultDumpThreads)
 	taskChannelCapacity.With(conf.Labels).Add(defaultDumpThreads)
-	var wg errgroup.Group
-	writers, tearDownWriters, err := d.startWriters(&wg, taskChan, rebuildConn)
+	wg, writingCtx := errgroup.WithContext(ctx)
+	writers, tearDownWriters, err := d.startWriters(writingCtx, wg, taskChan, rebuildConn)
 	if err != nil {
 		return err
 	}
@@ -203,9 +203,9 @@ func (d *Dumper) Dump() (err error) {
 	return nil
 }
 
-func (d *Dumper) startWriters(wg *errgroup.Group, taskChan <-chan Task,
+func (d *Dumper) startWriters(ctx context.Context, wg *errgroup.Group, taskChan <-chan Task,
 	rebuildConnFn func(*sql.Conn) (*sql.Conn, error)) ([]*Writer, func(), error) {
-	ctx, conf, pool := d.ctx, d.conf, d.dbHandle
+	conf, pool := d.conf, d.dbHandle
 	writers := make([]*Writer, conf.Threads)
 	for i := 0; i < conf.Threads; i++ {
 		conn, err := createConnWithConsistency(ctx, pool)
--- a/v4/export/dump.go
+++ b/v4/export/dump.go
@@ -1,3 +1,5 @@
+// Copyright 2020 PingCAP, Inc. Licensed under Apache-2.0.
+
 package export
 
 import (
--- a/v4/export/dump.go
+++ b/v4/export/dump.go
@@ -1,6 +1,7 @@
 package export
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"fmt"
@@ -311,9 +312,6 @@ func (d *Dumper) concurrentDumpTable(conn *sql.Conn, meta TableMeta, ir chan<- T
 	if err != nil {
 		return err
 	}
-	if selectField == "" {
-		selectField = "''"
-	}
 
 	orderByClause, err := buildOrderByClause(conf, conn, db, tbl)
 	if err != nil {
@@ -379,30 +377,16 @@ func (d *Dumper) concurrentDumpTiDBTables(conn *sql.Conn, meta TableMeta, ir cha
 	ctx, conf := d.ctx, d.conf
 	db, tbl := meta.DatabaseName(), meta.TableName()
 
-	pkNames, pkVals, err := selectTiDBTableSample(db, tbl, conn, meta)
-
+	handleColNames, handleVals, err := selectTiDBTableSample(db, tbl, conn, meta)
+	if len(handleVals) == 0 {
+		return nil
+	}
 	selectField, selectLen, err := buildSelectField(conn, db, tbl, conf.CompleteInsert)
 	if err != nil {
 		return err
 	}
-	if selectField == "" {
-		selectField = "''"
-	}
-
-	where := make([]string, 0, len(pkVals)+1)
-	where = append(where, fmt.Sprintf("(%s) < %s", pkNames, pkVals[0]))
-	for i := 1; i < len(pkVals); i++ {
-		low, up := pkVals[i-1], pkVals[i]
-		where = append(where, fmt.Sprintf("(%s) < (%s) AND (%s) >= (%s)", pkNames, low, pkNames, up))
-	}
-	where = append(where, fmt.Sprintf("(%s) >= (%s)", pkNames, pkVals[len(pkVals)-1]))
-
-	var orderByClause string
-	if pkNames == "_tidb_rowid" {
-		orderByClause = "ORDER BY `_tidb_rowid`"
-	} else {
-		orderByClause = fmt.Sprintf("ORDER BY %s", pkNames)
-	}
+	where := buildWhereClauses(handleColNames, handleVals)
+	orderByClause := buildOrderByClauseString(handleColNames)
 
 	for _, w := range where {
 		query := buildSelectQuery(db, tbl, selectField, buildWhereCondition(conf, w), orderByClause)
@@ -417,6 +401,62 @@ func (d *Dumper) concurrentDumpTiDBTables(conn *sql.Conn, meta TableMeta, ir cha
 		}
 	}
 	return nil
+}
+
+func selectTiDBTableSample(dbName, tableName string, db *sql.Conn, meta TableMeta) (handleCols []string, pkVals []string, err error) {
+	pkFields, err := GetPrimaryKeyColumns(db, dbName, tableName)
+	if err != nil {
+		return nil, nil, errors.Trace(err)
+	}
+	query := buildTiDBTableSampleQuery(pkFields, dbName, tableName)
+	handleCols, handleColTypes := findHandleColsAndColTypes(pkFields, meta)
+	rows, err := db.QueryContext(context.Background(), query)
+	if err != nil {
+		return nil, nil, errors.Trace(err)
+	}
+	iter := newRowIter(rows, len(handleCols))
+	defer iter.Close()
+	rowRec := MakeRowReceiver(handleColTypes)
+	buf := new(bytes.Buffer)
+	for iter.HasNext() {
+		err = iter.Decode(rowRec)
+		if err != nil {
+			return nil, nil, errors.Trace(err)
+		}
+		rowRec.WriteToBuffer(buf, true)
+		pkVals = append(pkVals, buf.String())
+		buf.Reset()
+		iter.Next()
+	}
+	return handleCols, pkVals, nil
+}
+
+func buildTiDBTableSampleQuery(pkFields []string, dbName, tblName string) string {
+	hasImplicitRowID := pkFields == nil
+	if hasImplicitRowID {
+		template := "SELECT `_tidb_rowid` FROM `%s`.`%s` TABLESAMPLE REGIONS() ORDER BY `_tidb_rowid`"
+		return fmt.Sprintf(template, dbName, tblName)
+	}
+	template := "SELECT %s FROM `%s`.`%s` TABLESAMPLE REGIONS() ORDER BY %s"
+	return fmt.Sprintf(template, pkFields, dbName, tblName, pkFields)
+}
+
+func findHandleColsAndColTypes(pkFields []string, meta TableMeta) (handleCols []string, colTypes []string) {
+	hasImplicitRowID := pkFields == nil
+	if hasImplicitRowID {
+		return []string{"_tidb_rowid"}, []string{"BIGINT"}
+	}
+	// Find primary key column types in meta.
+	colTypes = make([]string, len(pkFields))
+	for i, pkColName := range pkFields {
+		for j, colName := range meta.ColumnNames() {
+			if pkColName == colName {
+				colTypes[i] = meta.ColumnTypes()[j]
+				break
+			}
+		}
+	}
+	return pkFields, colTypes
 }
 
 func prepareTableListToDump(conf *Config, pool *sql.Conn) error {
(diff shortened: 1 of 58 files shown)


kennytm:
(blocked by pingcap/tidb#20883)

v4/export/dump.go
```diff
@@ -117,6 +118,11 @@ func (d *Dumper) Dump() (err error) {
 		return err
 	}
 	defer connectPool.Close()
+	writerConnPool, err := newConnectionsPool(ctx, conf.Threads, pool)
```
lichunzhu, line 121: I don't think we need two connection pools to parallelize splitting chunks and writing table data. It will create too many MySQL connections. In my test split chunks and dump these databases' and tables' meta SQLs usually cost less than 1s (around 300ms). The time cost to dump each tables' data is much larger than this.

If we need to parallelize this, I think an extra connection is enough.

v4/export/dump.go
```diff
@@ -196,76 +191,237 @@ func (d *Dumper) Dump() (err error) {
...
-			}, newDumpChunkBackoffer(canRebuildConn(conf.Consistency, conf.TransactionalConsistency)))
-		})
-	}
-	if err := g.Wait(); err != nil {
+	if err := dumpingGroup.Wait(); err != nil {
 		summary.CollectFailureUnit("dump", err)
 		return err
+	}
+	if err := writingGroup.Wait(); err != nil {
+		summary.CollectFailureUnit("dump", err)
```
lichunzhu, line 252:
```suggestion
		summary.CollectFailureUnit("dump table data", err)
```

v4/export/dump.go
```diff
@@ -196,76 +191,237 @@ func (d *Dumper) Dump() (err error) {
...
 
-func dumpDatabases(pCtx context.Context, conf *Config, connectPool *connectionsPool, writer Writer, rebuildConnFunc func(*sql.Conn) (*sql.Conn, error)) error {
+func (d *Dumper) dumpDatabases(connectPool *connectionsPool, writer *Writer) error {
+	ctx, conf := d.ctx, d.conf
 	allTables := conf.Tables
-	g, ctx := errgroup.WithContext(pCtx)
-	tableDataIRTotal := make([]TableDataIR, 0, len(allTables))
-	splitChunkStart := time.Now()
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithCancel(ctx)
```
lichunzhu, line 198: I think this looks clearer.
```suggestion
	conf := d.conf
	allTables := conf.Tables
	ctx, cancel := context.WithCancel(d.ctx)
```

v4/export/dump.go
```diff
@@ -196,76 +191,237 @@ func (d *Dumper) Dump() (err error) {
...
-				if err != nil {
-					return
-				}
-				return writer.WriteTableData(ctx, tableIR)
-			}, newDumpChunkBackoffer(canRebuildConn(conf.Consistency, conf.TransactionalConsistency)))
-		})
-	}
-	if err := g.Wait(); err != nil {
+	if err := dumpingGroup.Wait(); err != nil {
 		summary.CollectFailureUnit("dump", err)
```
lichunzhu, line 248:
```suggestion
		summary.CollectFailureUnit("dump meta data", err)
```

v4/export/writer.go
```diff
@@ -3,7 +3,9 @@ package export
 import (
 	"bytes"
 	"context"
+	"database/sql"
 	"fmt"
+	"github.com/pingcap/br/pkg/utils"
 	"strings"
 	"text/template"
```
lichunzhu, line 11:
```suggestion
	"strings"
	"text/template"

	"github.com/pingcap/br/pkg/utils"
```

v4/export/writer_util.go
```diff
@@ -528,3 +526,49 @@ func compressFileSuffix(compressType storage.CompressType) string {
...
+	FileFormatUnknown FileFormat = iota
+	FileFormatSQLText
+	FileFormatCSV
+)
+
+// String implement Stringer.String method.
+func (f FileFormat) String() string {
+	switch f {
+	case FileFormatSQLText:
+		return "text"
```
lichunzhu, line 543: Why do we prefer `text` rather than `sql` here?

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+			} else {
+				err = writer.WriteTableMeta(ctx, dbName, table.Name, meta.ShowCreateTable())
+			}
+			if err != nil {
+				return err
+			}
+
+			tableIRStream := make(chan TableDataIR, defaultDumpThreads)
+			writingGroup.Go(func() error {
+				return writer.WriteTableData(ctx, meta, tableIRStream)
```
lichunzhu, line 228: We should use a global `tableIRStream` for `writingGroup` and start `goroutine` for each chunk (each mysql connection) like before. I think a good solution for `writingGroup` is to start `Threads`*worker `goroutines`. Each goroutine will use the same connection and try to get a chunk from the global `tableIRStream` and do the dumping work.

Imagining that we have a huge table. We will have only one `tableIRStream`, and in `writer.WriteTableData` the dump progress for each chunk is serialized not parallelized. The performance will sharply regress.

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+		template := "SELECT `_tidb_rowid` FROM `%s`.`%s` TABLESAMPLE REGIONS() ORDER BY `_tidb_rowid`"
+		return fmt.Sprintf(template, dbName, tblName)
+	}
+	template := "SELECT %s FROM `%s`.`%s` TABLESAMPLE REGIONS() ORDER BY %s"
+	quotaPk := make([]string, len(pkFields))
+	for i, s := range pkFields {
+		quotaPk[i] = fmt.Sprintf("`%s`", s)
+	}
+	pks := strings.Join(quotaPk, ",")
+	return fmt.Sprintf(template, pks, dbName, tblName, pks)
```
lichunzhu, line 461: We should also escape `dbName` and `tblName`.

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
 	}
 
-	return []TableDataIR{tableIR}, nil
-}
+	var colTypes []*sql.ColumnType
+	// If all columns are generated
+	if selectField == "" {
+		colTypes, err = GetColumnTypes(conn, "*", db, tbl)
+	} else {
+		colTypes, err = GetColumnTypes(conn, selectField, db, tbl)
```
lichunzhu, line 517: We should check the `err` here.

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+		if err != nil {
+			return meta, err
 		}
+		meta.showCreateTable = createTableSQL
+		meta.showCreateView = createViewSQL
+		return meta, nil
 	}
-	if err := g.Wait(); err != nil {
-		return true, chunksIterArray, err
+	createTableSQL, err := ShowCreateTable(conn, db, tbl)
```
lichunzhu, line 543: This `err` is not caught.

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+		case <-ctx.Done():
+			return true
+		case irChan <- ir:
+			return false
+		default:
+			sleepTime := 1 * time.Second
+			time.Sleep(sleepTime)
+			log.Warn("write channel is full, waiting writer to consume IR",
+				zap.Int("channel size", defaultDumpThreads),
+				zap.Stringer("sleep time", sleepTime))
```
lichunzhu, line 351: I think we don't need this 🤔 If we want to check how many ir are blocked here we can add a metrics and `Inc` before `select`.

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+}
+
+func (d *Dumper) sequentialDumpTable(conn *sql.Conn, meta TableMeta, ir chan<- TableDataIR) error {
+	conf := d.conf
+	db, tbl := meta.DatabaseName(), meta.TableName()
+	tableIR, err := SelectAllFromTable(conf, conn, db, tbl)
 	if err != nil {
 		return err
 	}
+	ir <- tableIR
```
lichunzhu, line 263: Should we pass `ctx` here and reuse `sendDataIRToChan`?

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+			time.Sleep(sleepTime)
+			log.Warn("write channel is full, waiting writer to consume IR",
+				zap.Int("channel size", defaultDumpThreads),
+				zap.Stringer("sleep time", sleepTime))
 		}
-		createTableSQL, err := ShowCreateTable(db, dbName, tableName)
+	}
+}
+
+var z = &big.Int{}
```
lichunzhu, line 356: Why do we use `z` as a global variable? 🤔

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+	hasImplicitRowID := pkFields == nil
+	if hasImplicitRowID {
+		return []string{"_tidb_rowid"}, []string{"BIGINT"}
+	}
+	// Find primary key column types in meta.
+	colTypes = make([]string, len(pkFields))
+	for i, pkColName := range pkFields {
+		for j, colName := range meta.ColumnNames() {
+			if pkColName == colName {
+				colTypes[i] = meta.ColumnTypes()[j]
```
lichunzhu, line 474: We will build a string array every time we use `meta.ColumnTypes`. It looks a little strange.
What about saving `columnNames` and `columnTypes` instead of `[]*sql.ColumnType` at first?

lichunzhu: What's more, I think we can select both primary key columns and their types through one SQL from MySQL/TiDB. I think that's a better implementation than this way.

v4/export/sql.go
```diff
@@ -537,11 +512,11 @@ func createConnWithConsistency(ctx context.Context, db *sql.DB) (*sql.Conn, erro
 	return conn, nil
 }
 
-func buildSelectField(db *sql.Conn, dbName, tableName string, completeInsert bool) (string, error) {
+func buildSelectField(db *sql.Conn, dbName, tableName string, completeInsert bool) (string, int, error) {
```
lichunzhu, line 515: Could you please help add a comment before this line to explain the return values of this function?

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+func (d *Dumper) concurrentDumpTiDBTables(conn *sql.Conn, meta TableMeta, ir chan<- TableDataIR) error {
+	ctx, conf := d.ctx, d.conf
+	db, tbl := meta.DatabaseName(), meta.TableName()
+
+	handleColNames, handleVals, err := selectTiDBTableSample(db, tbl, conn, meta)
+	if err != nil {
+		return err
+	}
+	if len(handleVals) == 0 {
+		return nil
```
lichunzhu, line 402: Should we return an error or some instructions here to help `sequentialDumpTable` or `concurrentDumpTable` using the unique index? This table will directly be skipped if we `return nil` here.

v4/export/test_util.go
```diff
@@ -86,6 +86,14 @@ type mockTableIR struct {
 	SQLRowIter
 }
 
+func (m *mockTableIR) ShowCreateTable() string {
+	panic("implement me")
```
lichunzhu, line 90: Just `return ""` here?

v4/export/dump.go
```diff
@@ -196,193 +188,386 @@ func (d *Dumper) Dump() (err error) {
...
+		rowRec.WriteToBuffer(buf, true)
+		pkVals = append(pkVals, buf.String())
+		buf.Reset()
+		iter.Next()
+	}
+	return handleCols, pkVals, nil
+}
+
+func buildTiDBTableSampleQuery(pkFields []string, dbName, tblName string) string {
+	hasImplicitRowID := pkFields == nil
```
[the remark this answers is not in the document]
tangenta: > Will all TiDB in v5.0.0 enable cluster index by default?
Yes. 

It is possible that a table have both `_tidb_rowid` and `pkFields`. Maybe we should check `_tidb_rowid` first here.

v4/export/dump.go
tangenta: =_= Perhaps we need to set a linter for dumpling.

lichunzhu: Yes... Maybe we can file another PR to do this.

v4/export/dump.go
tangenta: If the `handleVals` is empty, there is no data in this table.

v4/export/dump.go
```diff
@@ -114,164 +117,384 @@ func (d *Dumper) Dump() (err error) {
...
-		})
+	return nil
+}
+
+func (d *Dumper) dumpTableData(conn *sql.Conn, meta TableMeta, taskChan chan<- Task) error {
+	conf := d.conf
+	if conf.NoData {
+		return nil
+	}
+	if conf.Rows == UnspecifiedSize {
```
lichunzhu, line 280: Will we support specifying rows for `tablesample` in the future? If not, specifying useless `rows` to enable dumping through `tablesameple` looks weird.

lichunzhu (commented):
Rest LGTM.
Amazing job!

lichunzhu:
@kennytm PTAL

v4/export/dump.go
tangenta: Yes. Later we should also add some tests for this feature.

lichunzhu (commented):
LGTM

lichunzhu (approved):

PR #208, opened 2020-12-02
export: don't write SHOW MASTER STATUS when lacking privilege

lance6716:

part of https://github.com/pingcap/dm/issues/1320

don't write SHOW MASTER STATUS when lacking privilege

### Check List 

Tests 

 - Unit test
 - Manual test

Side effects

 - Breaking backward compatibility (**but keep same with mydumper**)

Related changes

 
### Release note

- No release note


lichunzhu (approved):
LGTM

PR #212, opened 2020-12-03
*: integrate golangci-lint for dumpling

lichunzhu:
Dumpling didn't have linter before. The code has some problems not detected before.

Integrate `golangci-lint` in dumpling and and `make check` in CI.

### Check List 

Tests 

 - Unit test
 - Integration test
 
### Release note

- No release note


Makefile
```diff
@@ -1,40 +1,120 @@
...
+tidy:
+	@echo "go mod tidy"
+	GO111MODULE=on go mod tidy
+	cd tools && GO111MODULE=on go mod tidy
+	git diff --exit-code go.mod go.sum tools/go.mod tools/go.sum
 
-failpoint-disable: bin/failpoint-ctl
-# Restoring gofail failpoints...
-	@$(FAILPOINT_DISABLE)
+.PHONY: tools
```
kennytm, line 120: .PHONY should include every target which is not on the filesystem...

v4/export/config.go
```diff
@@ -608,23 +628,20 @@ func (s ServerType) String() string {
...
-var serverTypeString []string
-
-func init() {
-	serverTypeString = make([]string, ServerTypeAll)
-	serverTypeString[ServerTypeUnknown] = "Unknown"
-	serverTypeString[ServerTypeMySQL] = "MySQL"
-	serverTypeString[ServerTypeMariaDB] = "MariaDB"
-	serverTypeString[ServerTypeTiDB] = "TiDB"
-}
+var serverTypeString = []string{"Unknown", "MySQL", "MariaDB", "TiDB"}
```
kennytm, line 644:
```suggestion
var serverTypeString = []string{
	ServerTypeUnknown: "Unknown",
	ServerTypeMySQL:   "MySQL",
	ServerTypeMariaDB: "MariaDB",
	ServerTypeTiDB:    "TiDB",
}
```

cmd/dumpling/main.go
```diff
@@ -68,13 +68,11 @@ func main() {
 		fmt.Printf("\ncreate dumper failed: %s\n", err.Error())
 		os.Exit(1)
 	}
-	defer dumper.Close()
```
kennytm, line 71: can this be removed?

lichunzhu: The is reported by go-lint. Run `defer dumper.Close()` after `os.Exit(1)`.
I think we don't need to we dumper to `Close` when we finish dumping or meet some error. So I delete this line.

kennytm: that's strange, because in normal case (`err == nil`) the `os.Exit(1)` will not be executed and thus the `defer` would be run.

lichunzhu:
```
cmd/dumpling/main.go:76:3: exitAfterDefer: os.Exit clutters `defer dumper.Close()` (gocritic)
                os.Exit(1)
```
This error is reported from `gocritic`.

kennytm: it looks like [the lint](https://go-critic.github.io/overview.html#exitAfterDefer-ref)'s purpose is to ensure the `defer` get executed, not removing it.

maybe change this instead to:

```go
func run() error {
    ...
    if err != nil {
        log.Error(...)
        return err
    }
    ...
}

func main() {
    if err := run(); err != nil {
        os.Exit(1)
    }
}
```

Makefile
lichunzhu: Yes. It sames that BR's makefile also need to be updated 🤣

kennytm: adding to .PHONY is not necessary (basically it just prevents a `./tools` file cause `make tools` do nothing) but at least keep the original ones 😓

lichunzhu: I didn't know the meaning of `.PHONY` before and directly copied this from BR. I have updated Makefile now.

lichunzhu:
@kennytm all comments have been addressed. PTAL again.

v4/export/prepare.go
```diff
@@ -105,53 +110,65 @@ func listAllViews(db *sql.Conn, databaseNames []string) (DatabaseTables, error)
...
 
+// AppendTables appends several basic tables to DatabaseTables
 func (d DatabaseTables) AppendTables(dbName string, tableNames ...string) DatabaseTables {
 	for _, t := range tableNames {
 		d[dbName] = append(d[dbName], &TableInfo{t, TableTypeBase})
 	}
 	return d
 }
 
+// AppendViews appends several view to DatabaseTables
```
tangenta, line 156:
```suggestion
// AppendViews appends several views to DatabaseTables
```

tangenta (commented):
LGTM

kennytm (commented):
lgtm

kennytm (approved):

PR #214, opened 2020-12-07
sql: fix the problem that specify --sql will query twice

tangenta:

In the previous implementation, using `--sql` option triggers query twice.

This PR add a new implementation for `TableMeta` and `TableDataIR` to avoid query twice.

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

N/A

Related changes

N/A
 
### Release note

- No release note


kennytm:
could you add a test case to ensure the query is only run once (maybe like dump `select nextval(some_seq)`)? rest LGTM

Issue #218, opened 2020-12-09
Dumpling's output csv file is not correct when we set csv-delimiter to ""


lichunzhu:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
Run `dumpling --filetype csv --csv-delimiter "" --csv-seperater ","`, the output csv is invalid.

```sql
insert into t values
('0,00'),
('0,,0');
```

dumpling's output file:

```csv
0,00
0,,0
```

2. What did you expect to see?

`dumpling` should output the right csv file.

```csv
0\,00
0\,\,00
```

lichunzhu:
Affected versions: [v4.0.2:v4.0.8]

lichunzhu:
**1. Root Cause Analysis (RCA)**

Dumpling assumes that sql values don't contain characters of `csv-separators` and didn't handle these characters. When users set `--filetype=csv --csv-delimiter=""`, the exported csv file will be invalid if any SQL value contains the whole separator string.

**2. Symptom**

The data will be exported successfully but can be imported into tidb cluster.

**3. All Trigger Conditions**

All of:

1. Using Dumpling to export CSV files
2. The setting `--csv-delimiter` is set to ""(empty).
3. The table to dump has string columns and the column value contains the whole `csv-separator`.

Example: Setting `--csv-delimiter="" --csv-separator="," --filetype csv`. The table to be exported is:
```sql
insert into t values
('0,00', '0,,0'),
('0,,0', 'a');
```
The exported csv will be:
```csv
0,00,0,,0
0,,0,a
```
It's an invalid csv file.

**4. Workaround**

If Dumpling was used to export CSV files, set `--csv-delimiter='"'` to assure safety. **DO NOT** use dumpling without `csv-delimiter`.

If the data is already exported, please set `--csv-delimiter` to a not empty value and export again.

**5. Affected versions**

[v4.0.2:v4.0.8]

**6. Fixed versions**

v4.0.9

PR #219, opened 2020-12-10
fix the bug that output csv files are invalid when csv-delimiter is empty

lichunzhu:
close https://github.com/pingcap/dumpling/issues/218
dumpling's output csv file is not correct when `csv-delimiter` is empty and the string value contains seprator.

Add special judge for `csv-delimiter` is empty.
Add e2e_csv integration test to test csv e2e correctness.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- fix the bug that output csv files are invalid when csv-delimiter is empty


v4/export/sql_type.go
```diff
@@ -101,6 +102,64 @@ func escape(s []byte, bf *bytes.Buffer, escapeQuotation []byte) {
...
+	)
+	if len(delimiter) > 0 {
+		bf.Write(bytes.ReplaceAll(s, delimiter, append(delimiter, delimiter...)))
+		return
+	}
+	if len(separator) > 0 {
+		separate = separator[0]
+	}
+
+	for i := 0; i < len(s); i++ {
```
kennytm, line 132: i don't think we should care when `escapeBackslash` is false (when the default is true). if the user explicitly provided `--csv-delimiter="" --escape-backslash=false`, it is *the user's fault* that the data contains a `csv-separator`.

this has the same behavior as `SELECT INTO OUTFILE`:

```sql
values row('foo,bar', 'baz'), row('foo', 'bar,baz') 
    into outfile '1.txt' 
    fields terminated by ',' escaped by '';
```

result:

```
foo,bar,baz
foo,bar,baz
```

lichunzhu: I see. Do we need to escape all the characters that SQL escaped? Or we just need to escape byte 0 like TiDB? https://github.com/pingcap/tidb/blob/c33e90a7aef495bb2e86311b4c62ee98f30ae864/executor/select_into.go#L98

kennytm: when `escapeBlackslash` is true, we must always escape 0, the line terminators, the delimiter, and the separator (when the delimiter is empty). the rest are just bonus (SELECT INTO does not escape them).

when `escapeBlackslash` is false, nothing should be escaped, not even 0.

kennytm:
```
Failed to parse command flags:  Cannot parse config file `/tmp/dumpling_test_result/conf/lightning.toml`: Near line 17 (last key parsed 'mydumper.csv.delimiter'): invalid escape character 'N'; only the following escape characters are allowed: \b, \t, \n, \f, \r, \", \\, \uXXXX, and \UXXXXXXXX
```

lichunzhu:
> e2e_csv test failed
> 
> ```
> Failed to parse command flags:  Cannot parse config file `/tmp/dumpling_test_result/conf/lightning.toml`: Near line 17 (last key parsed 'mydumper.csv.delimiter'): invalid escape character 'N'; only the following escape characters are allowed: \b, \t, \n, \f, \r, \", \\, \uXXXX, and \UXXXXXXXX
> ```

This CI's shell version seems not compatible with mine. These tests can pass on my own PC. 🤔 Let me take a check

lichunzhu:
@kennytm PTAL

tests/e2e_csv/run.sh
```diff
@@ -0,0 +1,82 @@
...
+    run_dumpling --filetype="csv" --escape-backslash=$escape_backslash --csv-delimiter="$csv_delimiter" --csv-separator="$csv_separator"
+
+    # construct lightning configuration
+    mkdir -p $DUMPLING_TEST_DIR/conf
+    cp "$cur/conf/lightning.toml" $DUMPLING_TEST_DIR/conf
+
+    sed -i -e "s/separator-place-holder/$csv_separator/g" $DUMPLING_TEST_DIR/conf/lightning.toml
+    csv_delimiter_holder=$csv_delimiter
+    if [ "$csv_delimiter" = '"' ]; then
+        csv_delimiter_holder="\\\\\\\""
```
kennytm, line 48:
```suggestion
        csv_delimiter_holder='\\\"'
```

perhaps also comment why so many backslashes are needed

v4/export/sql_type.go
```diff
@@ -101,6 +90,66 @@ func escape(s []byte, bf *bytes.Buffer, escapeQuotation []byte) {
...
+		escape = 0
+
+		switch s[i] {
+		case 0: /* Must be escaped for 'mysql' */
+			escape = '0'
+		case '\n': /* escaped for line terminators */
+			escape = 'n'
```
kennytm, line 112:
```suggestion
		case '\r':
			escape = 'r'
		case '\n': /* escaped for line terminators */
			escape = 'n'
```

v4/export/sql_type.go
```diff
@@ -101,6 +90,66 @@ func escape(s []byte, bf *bytes.Buffer, escapeQuotation []byte) {
...
+			bf.WriteByte('\\')
+			bf.WriteByte(escape)
+			last = i + 1
+		}
+	}
+	if last == 0 {
+		bf.Write(s)
+	} else if last < len(s) {
+		bf.Write(s[last:])
+	}
```
kennytm, line 130:
```suggestion
	bf.Write(s[last:])
```

all branches are equivalent since `s[0:] == s` and `s[len(s):] == ""`

v4/export/sql_type.go
```diff
@@ -101,6 +90,66 @@ func escape(s []byte, bf *bytes.Buffer, escapeQuotation []byte) {
...
+	} else {
+		bf.Write(bytes.ReplaceAll(s, quotationMark, twoQuotationMarks))
+	}
+}
+
+func escapeCSV(s []byte, bf *bytes.Buffer, escapeBackslash bool, opt *csvOption) { // revive:disable-line:flag-parameter
+	if escapeBackslash {
+		escapeBackslashCSV(s, bf, opt)
+	} else {
+		if len(opt.delimiter) > 0 {
```
kennytm, line 145:
```suggestion
	} else if len(opt.delimiter) > 0 {
```

kennytm (commented):
rest LGTM

overvenus:
LGTM

kennytm (approved):

PR #230, opened 2021-01-11
conf: support flagset in config

GMHDBJD:
integrate flagset into export.config, so other components like dm can use flagset in dumpling too.

### DIFF
--- a/cmd/dumpling/main.go
+++ b/cmd/dumpling/main.go
@@ -20,7 +20,6 @@ import (
 	"os"
 
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/spf13/pflag"
 	"go.uber.org/zap"
 
 	"github.com/pingcap/dumpling/v4/cli"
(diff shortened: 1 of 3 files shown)


lance6716:
Could you explain how DM will use `cfg.FlagSet`?

GMHDBJD:
> Could you explain how DM will use `cfg.FlagSet`?

DM can use conf.ParseAndAdjust for extra args now. We now no need to change DM if dumpling add new argument.

lance6716:
> > Could you explain how DM will use `cfg.FlagSet`?
> 
> DM can use conf.ParseAndAdjust for extra args now. We now no need to change DM if dumpling add new argument.

DM has saved some configuration in `subtaskCfg`, we still need to set them one by one? And will `ParseAndAdjust` overwrite some config to default value

GMHDBJD:
previous use
```
dumpConfig := export.DefaultConfig()
dumpConfig.Host = db.Host
...
dumplingFlagSet = pflag.NewFlagSet
dumplingFlagSet.StringSliceVarP
...
dumplingFlagSet.Parse(args)
```

now
```
dumpConfig := export.DefaultConfig()
dumpConfig.Host = db.Host
...
dumplingFlagSet.Parse(args)
```

So now DM no need to update FlagSet when dumpling add new config.

> ParseAndAdjust overwrite some config to default value

That is expected behavior. If user specify the extra args for some arg, they should be overwritten.

lance6716:
>>ParseAndAdjust overwrite some config to default value
>
>That is expected behavior. If user specify the extra args for some arg, they should be overwritten.

I mean 
```
dumpConfig := export.DefaultConfig()
dumpConfig.Host = db.Host
...
dumpConfig.ParseAndAdjust(args) // will reset host to flagset default value
```

GMHDBJD:
> will reset host to flagset default value

`flagset.Parse(args)` will not overwrite value if it's not specified in args.
Others like `filesize`, which we set the values by `FlagSet.GetString()` will be overwritten with default value if not specified. I think that makes sense.

lance6716:
> > will reset host to flagset default value
> 
> `flagset.Parse(args)` will not overwrite value if it's not specified in args.
> Others like `filesize`, which we set the values by `FlagSet.GetString()` will be overwritten with default value if not specified. I think that makes sense.

if user both set `filesize` in `chunk-filesize` and `extra-args`, and we promise `extra-args` has higher priority, then DM should first set `dumpConfig.FileSize` then `dumpConfig.ParseAndAdjust(args)`. 

But if user only set `chunk-filesize`, handling in above order will lost that value.

GMHDBJD:
> if user both set `filesize` in `chunk-filesize` and `extra-args`, and we promise `extra-args` has higher priority, then DM should first set `dumpConfig.FileSize` then `dumpConfig.ParseAndAdjust(args)`.

You're right. I close the pr.

lance6716:
we might find a way to let filesize behaves like host 🤣 will think of it later

PR #241, opened 2021-02-24
log: support concurrently call log's function

lance6716:

when using dumpling as library, data race may occure because there's reading and writing on global variable `appLogger`

atomic action

### Check List 

Tests 

 - Unit test

Side effects

Related changes

 
### Release note


lance6716:
replaced by #242

PR #242, opened 2021-02-25
log, metrics: refactor metrics and log system for dm and other repositories to use

lichunzhu:
1. Now dumpling only has only one global logger. DM may init several dumpling loggers, which may cause a data race in dumpling. Besides, each logger has a different task name, which may be confusing if we only have one global logger.
2. DM can't set customized `conf.Labels` because we have preset their labels in `metrics.go`.

1. Use a customized logger for each dumper instance.
2. Don't set prometheus vector at first.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- No release note


v4/context/context.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+	"time"
+
+	"github.com/pingcap/dumpling/v4/log"
+)
+
+// Context is used to in dm to record some context field like
+// * go context
+// * logger
+type Context struct {
```
lance6716, line 16: how about

```
type Context struct {
	context.Context
	logger log.Logger
}
```

so we don't need `(*Context).Context()`. and logger could only be accessed by `L()`

v4/export/metrics.go
```diff
@@ -10,61 +10,80 @@ import (
...
+	finishedRowsCounter            *prometheus.CounterVec
+	finishedTablesCounter          *prometheus.CounterVec
+	writeTimeHistogram             *prometheus.HistogramVec
+	receiveWriteChunkTimeHistogram *prometheus.HistogramVec
+	errorCount                     *prometheus.CounterVec
+	taskChannelCapacity            *prometheus.GaugeVec
+)
+
+// InitMetricsVector inits metrics vectors.
+// This function much run before RegisterMetrics
```
lance6716, line 23:
```suggestion
// This function must run before RegisterMetrics
```

v4/export/metrics.go
```diff
@@ -10,61 +10,80 @@ import (
...
+func InitMetricsVector(labels prometheus.Labels) {
```
lance6716, line 24: forget why we need change this 😂

v4/export/config.go
```diff
@@ -172,8 +172,8 @@ func DefaultConfig() *Config {
 // String returns dumpling's config in json format
 func (conf *Config) String() string {
 	cfg, err := json.Marshal(conf)
-	if err != nil {
-		log.Error("fail to marshal config to json", zap.Error(err))
+	if err != nil && conf.Logger != nil {
```
lance6716, line 175: didn't find where `conf.Logger` is set

lance6716 (commented):
rest LGTM

v4/export/config.go
lichunzhu: `conf.Logger` is set in DM, outside dumpling repository.

v4/context/context.go
lichunzhu: addressed in https://github.com/pingcap/dumpling/pull/242/commits/fcc2ab6e9f673aa3a09baea0d4cae588cd9fbaaa

lance6716:
/lgtm

need remove some call of `Context()` in test files

lance6716 (approved):

PR #254, opened 2021-03-25
add estimate total rows count to metrics for dump table

recall704:
add estimate total rows count

add prometheus metrics to save estimate_total_rows before you dump data

### Limitation
it does not work for `-S or --sql` and views data

### DIFF
--- a/v4/export/dump.go
+++ b/v4/export/dump.go
@@ -209,6 +209,12 @@ func (d *Dumper) Dump() (dumpErr error) {
 		}
 	})
 
+	// get estimate total count
+	err = d.getEstimateTotalRowsCount(tctx, metaConn)
+	if err != nil {
+		return err
+	}
+
 	if conf.SQL == "" {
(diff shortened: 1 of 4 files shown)


v4/export/dump.go
```diff
@@ -209,6 +209,12 @@ func (d *Dumper) Dump() (dumpErr error) {
 		}
 	})
 
+	// get estimate total count
+	err = d.getEstimateTotalRowsCount(tctx, metaConn)
+	if err != nil {
+		return err
```
lichunzhu, line 215: I think this isn't a fatal error. We can just `log.Error` here.

lichunzhu (commented):
rest LGTM

v4/export/dump.go
recall704: fixed

lichunzhu:
Could you please help sign the CLA? https://github.com/pingcap/dumpling/pull/254#issuecomment-806522513

Thanks a lot for your contribution!

Issue #243, opened 2021-03-08
dumpling may get blocked without any write when the connected database is lost


lichunzhu:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do? If possible, provide a recipe for reproducing the error.
Use dumpling to dump data from a TiDB instance. TiDB quitted because of OOM.

2. What did you expect to see?
Dumpling can report the error and quit ASAP after an appropriate retry.

3. What did you see instead?
Dumpling got blocked without any writing.

4. Versions of the cluster

    - Dumpling version (run `dumpling -V`):
       Newest master branch 2e8afecad6300f82d65a17d2176dbc15492a9cb9.

PR #256, opened 2021-03-27
dump: fix the problem that dumpling gets stuck after the database connection is lost

lichunzhu:
fix https://github.com/pingcap/dumpling/issues/243

Use `writerCtx` to avoid `taskChan` get blocked after the writer quitted. 

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### Release note

- fix the problem that dumpling gets stuck after the database connection is lost

### DIFF
--- a/v4/export/dump_test.go
+++ b/v4/export/dump_test.go
@@ -1,0 +1,56 @@
+// Copyright 2020 PingCAP, Inc. Licensed under Apache-2.0.
+
+package export
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/pingcap/errors"
+	"golang.org/x/sync/errgroup"
+
+	tcontext "github.com/pingcap/dumpling/v4/context"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	. "github.com/pingcap/check"
+)
+
+func (s *testSQLSuite) TestDumpBlock(c *C) {
+	db, mock, err := sqlmock.New()
+	c.Assert(err, IsNil)
+	defer db.Close()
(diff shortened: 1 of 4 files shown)


v4/export/sql_type.go
```diff
@@ -18,6 +18,7 @@ var (
 )
 
 func initColTypeRowReceiverMap() {
+	colTypeRowReceiverMap = map[string]func() RowReceiverStringer{}
```
overvenus, line 21: Dup with line 11?

overvenus (commented):
rest lgtm

v4/export/sql_type.go
lichunzhu: I want to clear this map every time we init this.

v4/export/dump_test.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2020 PingCAP, Inc. Licensed under Apache-2.0.
```
lance6716, line 1:
```suggestion
// Copyright 2021 PingCAP, Inc. Licensed under Apache-2.0.
```

v4/export/sql_type.go
lance6716: currently the only call of `initColTypeRowReceiverMap` is main routine 🤔 Or I guess that you could remove line 11

lance6716 (commented):
rest lgtm

v4/export/sql_type.go
lichunzhu, line 21:
```suggestion
```

PR #266, opened 2021-04-07
export: fix a metric not using labelNames

lance6716:
estimateTotalRowsCounter didn't respect labelNames, so panicked when using dumpling as a library and changing labelNames

### Check List 

Tests 

 - Manual test (WIP)

Side effects

Related changes
 
### Release note

- No release note


lance6716:
tested by a DM's integration test locally

PR #265, opened 2021-04-02
Add conn session vars

recall704:
### What problem does this PR solve? 

fix resetDBWithSessionParams if value is int or float

### What is changed and how it works?

try to parse `params` values string to int or float.

PR #262, opened 2021-04-01
sql: fix table sample sql generator bug

lichunzhu:
### What problem does this PR solve? 
https://github.com/pingcap/dumpling/pull/260 introduced some bugs.

### What is changed and how it works?
Fix them.

### Check List 

Tests 

 - Unit test
 - Integration test

### Release note

- No release note

PR #271, opened 2021-04-14
*: cherry-pick recent PRs for dumpling v5.0.1

lichunzhu:
Some features are not cherry-picked to release-5.0.

Cherry-pick PRs on the master branch to release-5.0.

BugFix:
- Fix the format problem for connection session variables. https://github.com/pingcap/dumpling/pull/265
- Fix panic bugs for metrics when used as library https://github.com/pingcap/dumpling/pull/266, https://github.com/pingcap/dumpling/pull/267
- Fix the problem that pd client's logs are not printed to dumpling's https://github.com/pingcap/dumpling/pull/270 

Feature:
- Add estimate total rows count to metrics for dump table. https://github.com/pingcap/dumpling/pull/254

Tests:
- Add more tests for sample SQL generator https://github.com/pingcap/dumpling/pull/262

### Check List 

Tests 

 - Unit test
 - Integration test
 
### Release note

- No release note

### DIFF
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -849,7 +849,12 @@ func pickupPossibleField(dbName, tableName string, db *sql.Conn, conf *Config) (
 }
 
 func estimateCount(tctx *tcontext.Context, dbName, tableName string, db *sql.Conn, field string, conf *Config) uint64 {
-	query := fmt.Sprintf("EXPLAIN SELECT `%s` FROM `%s`.`%s`", escapeString(field), escapeString(dbName), escapeString(tableName))
+	var query string
+	if strings.TrimSpace(field) == "*" || strings.TrimSpace(field) == "" {
+		query = fmt.Sprintf("EXPLAIN SELECT * FROM `%s`.`%s`", escapeString(dbName), escapeString(tableName))
+	} else {
+		query = fmt.Sprintf("EXPLAIN SELECT `%s` FROM `%s`.`%s`", escapeString(field), escapeString(dbName), escapeString(tableName))
+	}
 
 	if conf.Where != "" {
 		query += " WHERE "
--- a/v4/export/sql.go
+++ b/v4/export/sql.go
@@ -511,44 +511,23 @@ func isUnknownSystemVariableErr(err error) bool {
 	return strings.Contains(err.Error(), "Unknown system variable")
 }
 
-func resetDBWithSessionParams(tctx *tcontext.Context, db *sql.DB, dsn string, params map[string]interface{}) (*sql.DB, error) {
-	support := make(map[string]interface{})
-	for k, v := range params {
-		var pv interface{}
-		if str, ok := v.(string); ok {
-			if pvi, err := strconv.ParseInt(str, 10, 64); err == nil {
-				pv = pvi
-			} else if pvf, err := strconv.ParseFloat(str, 64); err == nil {
-				pv = pvf
-			} else {
-				pv = str
-			}
-		} else {
-			pv = v
-		}
-		s := fmt.Sprintf("SET SESSION %s = ?", k)
-		_, err := db.ExecContext(tctx, s, pv)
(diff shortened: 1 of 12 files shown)


v4/export/sql.go
```diff
@@ -514,8 +514,20 @@ func isUnknownSystemVariableErr(err error) bool {
 func resetDBWithSessionParams(tctx *tcontext.Context, db *sql.DB, dsn string, params map[string]interface{}) (*sql.DB, error) {
 	support := make(map[string]interface{})
 	for k, v := range params {
+		var pv interface{}
```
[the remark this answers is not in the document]
lance6716: reproduced error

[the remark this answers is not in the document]
lance6716: What if user input `'value'` not `value` for string type? I prefer use input directly to printf a SQL.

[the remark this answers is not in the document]
lance6716: I prefer we directly use input to construct a `SET` sql, not considering if they're numeric or not. "directly" means that we don't try infer the variable is string type and quote them, let user quote it if needed.

https://github.com/go-sql-driver/mysql/blob/bcc459a906419e2890a50fc2c99ea6dd927a88f2/connection.go#L68-L78

v4/export/sql.go
```diff
@@ -514,8 +514,20 @@ func isUnknownSystemVariableErr(err error) bool {
...
+				pv = pvi
+			} else if pvf, err := strconv.ParseFloat(str, 64); err == nil {
+				pv = pvf
+			} else {
+				pv = str
+			}
+		} else {
+			pv = v
+		}
 		s := fmt.Sprintf("SET SESSION %s = ?", k)
```
lance6716, line 529: Here using `?` will let sql driver to check the type of `pv`, to decide if `pv` need quotes.

Instead, I prefer always use the input value as-is, that is to say we always use `%s` and printf to construct a SQL. So user should quote the value if it's a string type variable in SQL.

https://github.com/go-sql-driver/mysql/blob/bcc459a906419e2890a50fc2c99ea6dd927a88f2/connection.go#L68-L78

lance6716: we might append `--param` to dsn 😂 otherwise we can't control https://github.com/go-sql-driver/mysql#parameters and have to write a doc about dealing with `,` in parameter KVs.

lichunzhu: > Here using `?` will let sql driver to check the type of `pv`, to decide if `pv` need quotes.
> 
> Instead, I prefer always use the input value as-is, that is to say we always use `%s` and printf to construct a SQL. So user should quote the value if it's a string type variable in SQL.
> 
> https://github.com/go-sql-driver/mysql/blob/bcc459a906419e2890a50fc2c99ea6dd927a88f2/connection.go#L68-L78
> 
> (I deleted previous misleading comment)

`conf.SessionParams` is a `map[string]interface{}`. I think using an original value to construct SQL is in-proper for being used as a library.

lance6716: what's the detailed in-proper case?

if user input `key=val`, successful parsing it as a integer does not mean that's an interger MySQL variable, at that time it will raise an error.

kennytm: I think that `SET X = 123` and `SET X = '123'` are equivalent

lichunzhu: @lance6716 @kennytm 
Fixed in https://github.com/pingcap/dumpling/pull/271/commits/1b33d20633bb27be2dfb1df4b069122b333a3cd5, and add an integration test in https://github.com/pingcap/dumpling/pull/271/commits/278a1289ac7825a963d2f0b86a8d857b9d40f889. PTAL again

lance6716: sorry I didn't reply to above comment in time 😂 I didn't found a case for system variables that `SET SESSION X = 123` will cause error if X is a string type variable. Will check it now.

PR #276, opened 2021-04-28
support dump with UTC timezone

glorv:
Support the `--tz-utc` config to dump timestamp value with UTC time_zone.

If set the `--tz-utc` command line parameter:
- Add comment `/*!40103 SET TIME_ZONE='+00:00' */;` for each schema and data files.
- Set `time_zone='+00:00'` parameter to the DSN 

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Side effects

Related changes

### DIFF
--- a/tests/basic/run.sh
+++ b/tests/basic/run.sh
@@ -26,6 +26,24 @@ cnt=$(grep -w "Welcome to dumpling.*Release Version.*Git Commit Hash.*Go Version
 echo "version info count is ${cnt}"
 [ "$cnt" = 1 ]
 
+# Test tz-utc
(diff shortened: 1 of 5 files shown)


kennytm:
why not use `--params time_zone=+00:00`

glorv:
> why not use `--params time_zone=+00:00`

I think `--params time_zone=+00:00` is ok too though it's a bit weird that we need to depend on the session variables to set the right file meta.

BTW, mydumper also support `--tz-utc` and `--skip-tz-utc`, @lichunzhu  what do you think?

lichunzhu:
I think `--params` is okay. Maybe we should ask advice from Ian? @glorv

PR #283, opened 2021-05-11
config: Correct usage info for snapshot flag

dveeden:

Make it clear that dumpling also accepts a MySQL timestamp format as string for the snapshot.

### Check List 

Tests 

 - No code

### Release note

- No release note


lichunzhu (commented):
/lgtm

PR #319, opened 2021-07-27
security: support init tls config from cert/key/ca bytes

Ehco1996:
as title

### Check List 

Tests 

 - Integration test (in raw tls test) 

Side effects

 - Increased code complexity

Related changes

### Release note

- No release note


Ehco1996:
@lichunzhu  ci make check timeout, but I don't have restart access, is there any command to do that?

v4/export/config.go
```diff
@@ -668,10 +673,32 @@ func adjustConfig(conf *Config, fns ...func(*Config) error) error {
...
+		tlsConfig, err = utils.ToTLSConfigWithVerifyByRawbytes(conf.Security.SSLCABytes,
+			conf.Security.SSLCertBytes, conf.Security.SSLKEYBytes, []string{})
 		if err != nil {
 			return errors.Trace(err)
 		}
+		// NOTE for local test(use a self-signed or invalid certificate), we don't need to check CA file.
+		// see more here https://github.com/go-sql-driver/mysql#tls
+		if conf.Host == "127.0.0.1" {
+			tlsConfig.InsecureSkipVerify = true
+		}
```
lichunzhu, line 701: Why ignore this here? Do we have examples in other programs?

Ehco1996: The self-signed certificate generated by `mysql_ssl_setup` does not contain 127.0.0.1

Then the go-mysql documentation also recommends skipping the ca check in this case

The example will be in the DM test

lichunzhu (approved):
LGTM

Issue #322, opened 2021-08-01
dump pending on `SHOW TABLE STATUS FROM all_generate_column`


Ehco1996:
## Question

i use `make integration_test` to run test

but dumpling pending on `begin to run Dump` for a long time

after check the pprof, i found that current goroutine is working on `ListAllDatabasesTables`

there is the  `show processlist` result

```
+-----+-----------------+-----------------+---------------------+---------+------+-------------------------+----------------------------------------------+
| Id  | User            | Host            | db                  | Command | Time | State                   | Info                                         |
+-----+-----------------+-----------------+---------------------+---------+------+-------------------------+----------------------------------------------+
| 4   | event_scheduler | localhost       | <null>              | Daemon  | 2086 | Waiting on empty queue  | <null>                                       |
| 114 | root            | localhost:54709 | all_generate_column | Sleep   | 750  |                         | <null>                                       |
| 115 | root            | localhost:54727 | <null>              | Query   | 0    | starting                | show full processlist                        |
| 160 | root            | localhost:55436 | <null>              | Sleep   | 93   |                         | <null>                                       |
| 161 | root            | localhost:55437 | <null>              | Sleep   | 93   |                         | <null>                                       |
| 162 | root            | localhost:55438 | <null>              | Query   | 93   | Waiting for commit lock | SHOW TABLE STATUS FROM `all_generate_column` |
+-----+-----------------+-----------------+---------------------+---------+------+-------------------------+----------------------------------------------+
```

and i found that mysql is wait `Waiting for commit lock`

### dumpling log

```
****************** Running test tests/all_generate_column/run.sh...
[2021年 8月 1日 星期日 14时37分00秒 CST] Executing SQL: drop database if exists all_generate_column;
[2021年 8月 1日 星期日 14时37分00秒 CST] Executing SQL: drop database if exists all_generate_column;
[2021年 8月 1日 星期日 14时37分00秒 CST] Executing SQL: create database all_generate_column;
[2021年 8月 1日 星期日 14时37分00秒 CST] Executing SQL: create table all_generate_column.t(a int as (1), b int as (2)) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
[...]

PR #323, opened 2021-08-02
*: use different connetion in consistencyTypeFlush

Ehco1996:

https://github.com/pingcap/dumpling/issues/322

use different mysql connetion between `FLUSH TABLES WITH READ LOCK` and `unlock tables`
 

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity

Related changes

### Release note

- fix pending on dump when use  mysql 8.0.19


Ehco1996:
If two connections are used to lock mysql's read lock will fail because mysql will implicitly execute `unlock tables` when the connection is closed for the statement `FLush TABLES WITH READ LOCK`.

PR #350, opened 2021-09-02
add a func for sort pkVals

docsir:
For dump.go,I have added a func for sort pkVals.
<BLANK LINE>
This change will make sure us to see dumpling dump data correctly.
<BLANK LINE>


lichunzhu (commented):
We shouldn't use string to save and sort all the values.
For integer values we may get "115 < 12" if we sort in this way.

